PR #164, opened 2016-10-26
Fixed method for adding new member into database

number213:
I noticed that `pixiv_master_member` table in my database is empty so I checked the code and found that DB manager's method `insertNewMember` is never called. So I fixed it and added code for calling it.

### DIFF
--- a/PixivDBManager.py
+++ b/PixivDBManager.py
@@ -286,21 +286,12 @@ class PixivDBManager:
 ##########################################
 ## IV. CRUD Member Table                ##
 ##########################################
-    def insertNewMember(self):
+    def insertNewMember(self, member_id, member_name):
         try:
             c = self.conn.cursor()
-            member_id = 0
-            while True:
-                temp = raw_input('Member ID: ')
-                try:
-                    member_id = int(temp)
-                except:
-                    pass
-                if member_id > 0:
-                    break
-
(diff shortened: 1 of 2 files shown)


Nandaka:
it is on purpose, so using option will not add new member. 

you use option 4 for that (it will import the member id from list.txt) and download the images from member id listed in DB.

sometimes I want to check new artist, but dont want to update the images from option 4.

number213:
So when you use option 1 and download by member id you don't need to write member's info into database?

Nandaka:
yep, i dont want to track the member id from option 1

number213:
Ok, got it.

jwshields:
I apologize for resurrecting such an old issue..
I also saw #465 and #453 about this as well.

May I ask the rationale behind `dont want to track the member id from option 1`?
For myself, that's something I would like to have, just to populate the table and keep track of artists.

Could this maybe be implemented if a flag was added to settings like `dbsavememberid = True` or something of that nature? If so, I could submit a PR-

Nandaka:
more because personal reason:

> it is on purpose, so using option will not add new member.
> 
> you use option 4 for that (it will import the member id from list.txt) and download the images from member id listed in DB.
> 
> sometimes I want to check new artist, but dont want to update the images from option 4.

jwshields:
I understand that from reading the comments, but my question was more of, is this could be implemented in a PR **with an option to enable or disable it**, would that be acceptable? This is something I would like _personally_ and based on the previous issues, others may like it too.

Nandaka:
sure, send me PR and I will review the code first.

PR #212, opened 2017-03-13
Made setup.py more colorful and detected python 3

awiebe:
Colorized terminal output for setup and made it check if it was run with python 3, so people running

`python setup.py`

Where the default interpreter is not 2.x will be warned they need to run using verison 2.


Nandaka:
can you pull latest commit and resend?

PR #261, opened 2017-09-16
fix r18mode url

srndpty:
r18mode url now should be `&mode=r18` to work.


NHOrus:
Does this include R18G now?

srndpty:
Perhaps. But I don't care about that since I have excluded it by adding `R-18G` in `blacklist_tags.txt`.

PR #296, opened 2018-01-03
Started agrssive PEP8 reformatting, whitespace and local variable names.

awiebe:
I've started trying to make the naming conventions in the project consistent with:
https://www.python.org/dev/peps/pep-0008/

So far I've just run over the code with pyCharm and renamed a number of local variables to use the python standard ``variable_name`` syntax.

I haven't changed function names yet

> Function names should be lowercase, with words separated by underscores as necessary to improve readability.
> 
> mixedCase is allowed only in contexts where that's already the prevailing style (e.g. threading.py), to retain backwards compatibility.

I'm not sure whether to go with mixed case everywhere or convert to underscored lowercase.  What's important is that the naming conventions are consistent in modules, and preferably consistent across PixivUtil.


Nandaka:
addheaders is from the mechanize.Browers libs, you shouldn't change it.

PixivBrowserFactory.py
```diff
@@ -79,20 +79,20 @@ def _configureBrowser(self, config):
             PixivHelper.GetLogger().info('Debug HTTP enabled.')
 
         # self.visit_response
-        self.addheaders = [('User-agent', config.useragent)]
```
Nandaka: should not be changed, as it is from the mechanize library

Issue #286, opened 2017-12-12
DB in use error when one process finishes if multiple processes are open


LieXXX:
If I use more than one PixivUtil2 process to download from Pixiv and one process finishes it's task. The other processes get a DB in use error. The finished process seems to lock the database file, preventing anything else from being downloaded.

EDIT: It does not seem to always happen.

Nandaka:
It is not supported by sqlite, as it doesn't support concurrent DB write operation from different process.

LieXXX:
I have been able to do multiple processes and they error but then continue when they get access. The issue is that when a process finishes and is asking me for input, it locks the DB for the other processes.

Nandaka:
can you try to append -x on the command?

anyway, running multiple instance from the same folder is not supported

LieXXX:
It runs so well from the same folder so well except for this one issue. The only thing I had to do was to increase the amount of retries when it errors out.

Where do you want me to add the -x command?

Nandaka:
you can use it if you are not running the application in interactive mode (e.g. from command line)

refer to https://github.com/Nandaka/PixivUtil2/blob/master/readme.txt#L178

PR #303, opened 2018-01-10
Increased SQLITE timeout

awiebe:
Sqlite has some issues with concurrent reads.  I don't know the exact nature but in analysing PixivUtil's behaviour, accessing the database with another sqlite tool, sometimes causes runtime exceptions(particularly for lengthy queries).  This seems to be because concurrent access tends to lock database pages which pixiv util also wants to update, causing the database connection timeout to pop.  I've hardcoded a timeout to 5 minutes.  In the future I suppose this could be added to config, but I don't see any reason why someone would need to allow for locks more than 5 minutes long.


NHOrus:
Why you are accessing DB with another tool while also running PixivUtil? This isn't supported by PixivUtil. Or SQLITE.

awiebe:
Actually SQLITE supports concurrent access just fine, it just can't handle concurrent write.
https://www.sqlite.org/lockingv3.html

I have been using http://sqlitebrowser.org/ to diagnose the contents of my database while waiting for a particularly long PixivUtil download task to complete.(I'm migrating my configuration, and was trying to figure out what was wrong with changing roots of manga images).  For some reason even though I'm only performing read only access, this seems to lock the database, this is either the problem of the GUI tool, or the SQLITE pager module implementation on HFS+, but either way, it goes away if I just bump up the timeout.

It has been mentioned that sqlite doesn't support concurrent write, https://github.com/Nandaka/PixivUtil2/issues/286  but again, this just means that only one process can update at a time,  as long as the timeout is long enough SQLITE can wait on other running processes.

I understand two instances of pixivutil for the same username at the same time risks clobbering data, since pixiv util doesn't have a lock to prevent that, but there is significant speed up running multiple simultaneous instances for different users ( which does not risk this because they are not accessing the same data files).

PixivDBManager.py
```diff
@@ -18,14 +18,14 @@ class PixivDBManager:
     """Pixiv Database Manager"""
     __config__ = None
 
-    def __init__(self, target='', config=None):
+    def __init__(self, target='', config=None,timeout=5*3600):
```
Nandaka: 5 minutes = 5 * 60, right?

the timeout is in second, right? meaning it is 5 hours?

Nandaka:
You are hardcoding the timeout to 5 hours (5 * 3600s)

PixivDBManager.py
awiebe: Ahm you're right, my mistake, 5*60 would be far more sane.

awiebe:
I have modified to be 5 minutes.

Issue #325, opened 2018-02-25
use python3 print format


rachmadaniHaryono:
hi, i'm trying to fork the program to python3.

download a single image was actually successful till this commit https://github.com/rachmadaniHaryono/PixivUtil2/commit/edcf06f301674f69799ac7f61d2705a782335e26

but with new update it is currently hard to track the change as the `print` is still mostly use python2 format

so i recommend `print` should be used as following

```python
# OLD
print 'Some text'
# NEW
print('Some text')
```
and `print` with trailing comma
```python
#OLD
print 'Some text',
#NEW
print('Some text', end='')
```
if you are okay with this, i can even create pr which consist only change for this

related #190

Nandaka:
Any impact with the old 2.7x version? Looks like print and print() has different behaviour, especially handling unicode string.
Also, please note that python2.7 has different handling if you pass more than one parameter.
See below
```
Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:42:59) [MSC v.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> print "hello", "world"
hello world
>>> print("hello", "world")
('hello', 'world')
>>>
```

rachmadaniHaryono:
not quite sure why the newest commit is not registered on pr, but here is the fix

https://github.com/rachmadaniHaryono/PixivUtil2/commit/881b85e54a54cf2109b2d5a40eee823bf96b8fca

rachmadaniHaryono:
result 

```python
Python 2.7.14 (default, Sep 23 2017, 22:06:14)                                      
[GCC 7.2.0] on linux2                     
Type "help", "copyright", "credits" or "license" for more information.              
>>> print "hello", "world"                
hello world                               
>>> print("hello", "world")               
('hello', 'world')                        
>>> from __future__ import print_function 
>>> print("hello", "world")                                                                                                                                             
hello world
```

rachmadaniHaryono:
another test for print for trailing comma and future lib

```python
#Python 2.7.14 (default, Sep 23 2017, 22:06:14)
>>> print 'hello',
hello
>>> print('hello',)
('hello',)
>>> from __future__ import print_function
>>> print('hello',)
hello
>>>

#Python 3.6.3 (default, Oct  3 2017, 21:45:48)
>>> print('hello',)
hello

>>> print('hello', end='')
hello
>>>
```

[...]

PR #326, opened 2018-02-25
use python3 print format

rachmadaniHaryono:
related #325 

this is not included fix for `print` with trailing space.

based on this https://stackoverflow.com/questions/4009672/python-trailing-comma-after-print-executes-next-instruction

there is no python2/3 compatibilities code, so the fix is probably using this format

```python
if six.PY3:
    print('Some text', end='')
else:
    print 'Some text',
```

and because of that fix, it require six library (which later can be useful to port to python3 code).

but for now because it add additional dependencies, i will put it on hold

### DIFF
--- a/PixivBrowserFactory.py
+++ b/PixivBrowserFactory.py
@@ -128,7 +128,7 @@ class PixivBrowser(mechanize.Browser):
                     for t in range(1, self._config.retryWait):
                         print t,
                         time.sleep(1)
-                    print ''
+                    print('')
                     retry_count = retry_count + 1
                 else:
                     raise PixivException("Failed to get page: " + ex.message, errorCode=PixivException.SERVER_ERROR)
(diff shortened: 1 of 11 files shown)


Nandaka:
You cannot simply add `()` if there are multiple `,`. Maybe refactor to use string.format()?
```
Python 2.7.13 (v2.7.13:a06454b1afa1, Dec 17 2016, 20:42:59) [MSC v.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> string = "test"
>>> print "hello", string
hello test
>>> print("hello", string)
('hello', 'test')
>>> print "hello", "world"
hello world
>>> print("hello", "world")
('hello', 'world')
>>>
```

rachmadaniHaryono:
based on this it can be fixed with single line

```python
from __future__ import print_function
```

Nandaka:
Looks like that fix the issue, another one: you need to use `end=' '` so it will append space for the continuation.

```
>>> if True:
...   print("Hello", end='')
...   print("World")
...
HelloWorld
>>> if True:
...   print("Hello", end=' ')
...   print("World")
...
Hello World
>>>
```

rachmadaniHaryono:
https://github.com/rachmadaniHaryono/PixivUtil2/commit/fd1fa52c59010a048e476b5f9b139893a909b680

not sure why the pr is not updating itself

Nandaka:
I think because I've closed the old one, try to raise another PR?

PR #328, opened 2018-02-25
Feature/python3 print format

rachmadaniHaryono:
Related #325


PixivHelper.py
```diff
@@ -816,7 +817,7 @@ def ugoira2webm(ugoira_file,
             chatter += buff
             if buff.endswith("\r"):
                 if chatter.find("frame=") > 0:
-                    print chatter.strip(), os.linesep,
+                    print(chatter.strip(), os.linesep,)
```
Nandaka: missed `, end=' '` here

supposed to be `print(chatter.strip(), os.linesep, end=' ')`, right?

PR #335, opened 2018-03-01
setup for non windows

rachmadaniHaryono:
with this it is possible to pip install the program, or even if you want upload it to pypi for easier deployment.

changes:

- prefer setuptools instead of disutils, reference: https://stackoverflow.com/questions/25337706/setuptools-vs-distutils-why-is-distutils-still-a-thing
- on non windows, the program will be run with `PixivUtil2`. it can be changed e.g. `PixivUtil2.py`, but i remove the `.py` extension to differentiate that it is command line.
- required package is looked up from `requirements.txt`   file. the reading method is taken from the reading method for program version. this is not foolproof. if the project require requirement, which starting with `git+` it have to be changed.
- method to get program version is taken from here https://stackoverflow.com/a/24517154/1766261
- program version is tried to use semver version. it mean `v20180218` will be `20180218.0.0` and `20180227-beta3` will be `20180227.0.0.beta3`
- no easy way for readme file. pypi require rst file and i have to convert the current readme into rst . rendered version for pypi can be found here https://test.pypi.org/project/PixivUtil2/ . uploaded file can be found here https://github.com/rachmadaniHaryono/PixivUtil2/blob/feature/readme-rst/readme.rst
- i don't know your email so leave `author_email` empty
- argument on `setup.py` is ordered based on this https://github.com/pypa/sampleproject/blob/master/setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,9 +3,14 @@
 from __future__ import print_function
 
 from distutils.core import setup
-
+from os import path
 import os
 import sys
+try:
+    from setuptools import setup, convert_path, find_packages
+except ImportError:
+    from distutils.core import setup, find_packages
+    from distutils.util import convert_path
 
 isWindows = os.name is 'nt'
 ranWithPy3 = sys.version_info >= (3, 0)
@@ -24,14 +29,6 @@ class bcolors:
 
 
 if not isWindows:
-    print(bcolors.FAIL)
-    print("____________________________")
-    print("          ERROR           ")
-    print("----------------------------")
-
-    print("Setup.py is used to generate executable under Windows systems\n")
-    print("For non windows systems please run:\n\n\tpip install -r requirements.txt")
-    print(bcolors.ENDC)
     if ranWithPy3:
         print(bcolors.WARNING)
         print("Attention: PixivUtil2 is not yet compatible with Python 3.  You have run this script with Python 3.")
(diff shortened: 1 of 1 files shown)


Nandaka:
Got some warning when running the setup.py, is it normal?
```
C:\Python27\lib\distutils\dist.py:267: UserWarning: Unknown distribution option: 'project_urls'
  warnings.warn(msg)
C:\Python27\lib\site-packages\setuptools\dist.py:351: UserWarning: Normalizing '20180227.0.0.beta3' to '20180227.0.0b3'
  normalized_version,
warning: py2exe: Version Info will not be included:
  could not parse version number '20180227.0.0b3'
```

technically you can break the versioning to yyyy.mm.dd.[beta#] , as that is the structure that I use.

rachmadaniHaryono:
>technically you can break the versioning to yyyy.mm.dd.[beta#] , as that is the structure that I use.

OK, I will do this

rachmadaniHaryono:
1. first the change to use the format you recommend. the new format pass the semver version, but on beta/alpha etc distutils will give warning that the version will be normalized.

example: `Normalizing '2018.2.2.beta3' to '2018.2.2b3'`

i think it is acceptable. the only problem i can think of is that when on a single day, there is multiple release

2. the `project_urls` is actually exist on setuptools, but if distutils put warning, i will add simple check so only when setuptool used, `project_urls` will be used on setup.py

rachmadaniHaryono:
hi @Nandaka, thank you for merging this pr,

did you think it is possible to upload it to pypi?

if it is, it will need:

- converting readme to rst
- update changelog
- changing the version number (beta4?)

Nandaka:
Any benefit if I upload to pypi?

rachmadaniHaryono:
- Easier installation. User dont need to update the program with git, instead they can directly use pip. Just `pip install PixivUtil2`, setting your config and the program is good  to go.
- easier upgrade. `pip install -U PixivUtil2' and the program will be upgraded.
- easier to use with other programs. Other dev can put PixivUtil2 as requirements easier. No need to put github link directly as requirements, only program name and version.

Con

- more work. See 3 things that need to be done above. If the first uploads success, next uploads should be easier. If everything working correctly, `python setup.py sdist upload -r pypi` is another command that have to be run on each release.

NHOrus:
Note: Don't ever `pip install`, always `pip install --user` because messing with system files? Bad idea.

And I honestly don't find serious need for the pypi. Unlike, say, streamlink, pixivutil2 designed to work locally, with database and config file in current directory. This can produce some interesting user errors. Changes in pixivutil2 to accomodate that, honestly, aren't worth it.


Also, more work for Nandaka that he may not wish to do, including your three changes, changes to make pixivutil2 work when located in `~/bin` or whatever instead of it's own folder with it's own database and config and logs without spamming and littering that, need to learn pypi ecosystem.

rachmadaniHaryono:
>Note: Don't ever pip install, always pip install --user because messing with system files? Bad idea.

Agree with this.

>. Unlike, say, streamlink, pixivutil2 designed to work locally, with database and config file in current directory. This can produce some interesting user errors. Changes in pixivutil2 to accomodate that, honestly, aren't worth it.

I haven't checked it fully yet, but on my branch with working setup.py, the config and database is working correctly. But still testing (on alpha branch?) is needed.

>Also, more work for Nandaka that he may not wish to do, including your three changes, changes to make pixivutil2 work when located in ~/bin or whatever instead of it's own folder with it's own database and config and logs without spamming and littering that, need to learn pypi ecosystem

Updating changelog and change the version version number is actually what Nandaka need every release right? So Imo it is normal thing to do. Regarding rst readme I already made proposals on above post ( https://github.com/rachmadaniHaryono/PixivUtil2/blob/feature/readme-rst/readme.rst)

>changes to make pixivutil2 work when located in ~/bin or whatever instead of it's own folder with it's own database and config and logs without spamming and littering that, need to learn pypi ecosystem.

This is actually the job for this setup.py file. If dev are worried about littering on system, I recommend to use virtualenv

Nandaka:
I'm using windows, so those thing doesn't really apply to me.

Unless you are running from source, upgrading just a matter of replacing the old files.

rachmadaniHaryono:
>No, I got no time to learn pypi at all.

OK, I expect you know something that have to be learned when uploading package to pypi. 

>I think my opinion about this all is PixivUtil2 isn't a systems program...,  plus it's executive, it works well as a tarball. Or git clone.

Really don't understand with this. What I can catch is this you mean it is source only program

>it's local, highly coupled program that's quite unsuited to be used as a library,

Well, I'm trying to use it as library and this may help me a lot. Sure there is more to be desired, but not used the program as library is kinda... Pity/unfortunate (? not quite sure which word fit here) 

>Finally, want non-maintenance feature? Nandaka's always accepting donations.
? Does it mean the program is maintenance only now? Or the new non - maintenance feature requires donations? 

---

First I want to say, there is difference between the setup.py and pypi feature. Most discussion here just mention problem that may arise from setup.py 

I am ok and will withdraw if there is valid pypi - related reason, or just 'it is too much hassle to do `python setup.py sdist upload -r pypi `on each release' kinda reason

Nandaka:
I don't want to maintain any issue from pypi basically. Of course you can import some of the script and treated as library, but this is designed as console first application/script.

Heck, I'm doing this because I'm using the application myself. Any feature request might be implemented if it is make senses and I see the benefit, which very subjective. I don't mind other people to use/extend the capabilities, but most of the time I don't really want to touch some one else code (unless I'm using that feature).

rachmadaniHaryono:
>I don't want to maintain any issue from pypi

If you don't want to upload to maintain pypi, you can ask someone to do that. 

Unless you mean installation from setup.py which nhorus already list some of the potential problem

>. I don't mind other people to use/extend the capabilities, but most of the time I don't really want to touch some one else code (unless I'm using that feature).

It is understandable. 

Currently PixivUtil2 need better documentation to be used as library. Other problem maybe examples and that global variable

PR #501, opened 2019-06-16
Allow value -1 for number of pages: stop after you find an image you already downloaded.

RubenGarcia:
Allow value -1 for number of pages: stop after you find an image you already downloaded.
This is useful if your database is not correctly synchronized with your archives (e.g. when reinstalling PixivUtil2).

### DIFF
--- a/PixivUtil2.py
+++ b/PixivUtil2.py
@@ -371,10 +371,14 @@ def process_list(list_file_name=None, tags=None):
         raise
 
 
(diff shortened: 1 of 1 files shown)


Nandaka:
you can use `checkUpdatedLimit` in config.ini

```
checkUpdatedLimit     ==> Jump to the next member id if already see n-number of
                          previously downloaded images.
                          alwaysCheckFileSize must be set to False.
```

PR #532, opened 2019-09-21
Python 3 support

fake-name:
It passes all the tests, if nothing else.

Py2k support is dropped, because you shouldn't be using python 2 anyways.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,12 @@
 language: python
 python:
-  - "2.7"
+  - "3.5"
+  - "3.6"
+  - "3.7"
 # command to install dependencies
 install:
   - pip install -e .
-  - pip install pytest
+  - pip install nosetests
 # command to run tests
 script:
-  - pytest -v ./test_*
+  - python3 -m nose --exe
--- a/PixivBrowserFactory.py
+++ b/PixivBrowserFactory.py
@@ -1,19 +1,19 @@
 # -*- coding: utf-8 -*-
 # pylint: disable=W0603, C0325
-from __future__ import print_function
 
-import cookielib
-import httplib
+
+import http.cookiejar
+import http.client
 import json
 import mechanize
 import re
 import socket
 import sys
 import time
-import urllib
-import urllib2
-import urlparse
-from BeautifulSoup import BeautifulSoup
+import urllib.request, urllib.parse, urllib.error
+import urllib.request, urllib.error, urllib.parse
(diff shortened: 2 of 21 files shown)


fake-name:
Ok, yeah, I don't know if this actually works, since it seems like the tests are not comprehensive.

My interest here is using this as a module in a larger application. As such, I'm doing [a large scale refactor](https://github.com/fake-name/PixivUtil2/tree/modular) with the end goal of removing all the global state (it's EVERYWHERE) (I've been maintaining [my own scraper](https://github.com/fake-name/xA-Scraper/blob/master/xascraper/modules/px/pxScrape.py), but yours seems FAR more comprehensive). 

Basically, I want to be able to use the webpage parsing stuff, without using the database or CLI stuff. Right now, they're extremely tightly interwoven. 

I'm currently trying to simply make them modular, with the end goal of hopefully getting the restructured code merged in a future pull request. 

I'd *like* to be able to just use a git submodule for this project so I can track upstream (and not have to manually fix my own scraper periodically). On the other hand, if you're not interested in a structure cleanup, I can just start deleting stuff I don't need.

Nandaka:
I'll check it, but if you really want to use it as module, it might not that suitable as it is not really designed for that.

Maybe you want to check https://github.com/upbit/pixivpy instead? It does support py3. It even have async support https://github.com/Mikubill/pixivpy-async.

Nandaka:
Pulled your changes to different branch and it break compatibility with python2.7.
Tried to run with python3.7 and it still won't run even after installing the requirements and update it to use BeautifulSoup4.
Changed the read() to read().decode('utf-8), now it can login, but looks like mechanize.Browser().open() cannot handler urllib2.request.Request() object...

fake-name:
> it break compatibility with python2.7.

At this point, you shouldn't be using python 2 anyways.

> Changed the read() to read().decode('utf-8), now it can login, but looks like mechanize.Browser().open() cannot handler urllib2.request.Request() object...

Yeah. That's fixed further on in https://github.com/fake-name/PixivUtil2/tree/modular . Why are you using a custom opener in the first place? It seems like mechanize can make posts itself by just passing the postdata as the `data` param to `open()`.

Nandaka:
> custom opener

SOCKS proxy supports

Nandaka:
Implemented python3 branch here. Run only with python 3.7++

https://github.com/Nandaka/PixivUtil2/tree/python3_switch

PR #546, opened 2019-10-25
Add BOM automatically

### LINKED ISSUE #543: Irfan Viewer and unicode filenames
This is more of "What I am doing wrong with file lists" question

Lists of downloaded images open perfectly with Irfan 4.44 with unicode plagin, but modern Irfan 4.53 opens only files without japanese characters in names.
Moveover, when I open files through Irfan 4.44 "Open in editor" functionality, Japanese filenames turn into glibberish.

Any ideas?

NHOrus:
Should fix #543

### DIFF
--- a/PixivUtil2.py
+++ b/PixivUtil2.py
@@ -281,7 +281,7 @@ def download_image(url, filename, referer, overwrite, max_retry, backup_old_file
 
                 # write to downloaded lists
                 if start_iv or __config__.createDownloadLists:
-                    dfile = codecs.open(dfilename, 'a+', encoding='utf-8-sig')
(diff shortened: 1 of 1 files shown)


NHOrus:
No, this does for first line only.

NHOrus:
Fixed

Nandaka:
won't this impact other viewer which doesn't support BOM or older irfanView? might need to add config to enable this?

NHOrus:
And this code doesn't work, because I miswrote getsize function.
Will make new PR later.

Old Irfan supports BOM, I tested. So, no problems there.

Issue #548, opened 2019-10-31
UnicodeEncodeError


DukeValentine:
### Prerequisites

* [ x] Did you read FAQ section?
* [x ] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?

### Description

So far every attempt to download anything has causes errors similar to this:

```

Traceback (most recent call last):
  File "PixivUtil2.py", line 176, in download_image
    if os.path.exists(filename) and os.path.isfile(filename):
  File "/usr/lib/python2.7/genericpath.py", line 26, in exists
    os.stat(path)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 26-38: ordinal not in range(128)
Checking local filename... Error at download_image(): (<type 'exceptions.UnicodeEncodeError'>, UnicodeEncodeError('ascii', u'/home/humanix/files/pixiv/\u3048\u3063\u3061\u3043\u306e\u306f\u597d\u304d\u3067\u3059\u304b\uff1f\uff1f (43636871)/77541148_p0 - \u60aa\u9b54\u30b3\u30b9\u306e\u5bfe\u9b54\u5fcd\u3055\u3093.png', 26, 39, 'ordinal not in range(128)'), <traceback object at 0x7f41f43b6cf8>) at https://i.pximg.net/img-original/img/2019/10/29/21/26/48/77541148_p0.png (9000)
```

### Versions

PixivDownloader2 version 20190919
OS: Ubuntu 16.04 LTS

Nandaka:
try to set the python encoding to utf-8 in console?
`export PYTHONIOENCODING=UTF-8`

DukeValentine:
It works fine on another machine, so I really don't know what's going on. Are you sure there isn't anything else that must be installed in order for this application to work?

Nandaka:
not sure, maybe permission issue? I don't use linux so I'm not familiar with it.

DukeValentine:
No, it couldn't be a permission issue. It's, as the message shows, an encoding issue

DukeValentine:
What seems to mitigate the issue is using filename.encode('utf-8') in all places that cause this kind of exception. 
Weird that this only happens in this specific machine, I wonder if there's a config issue. But  export PYTHONIOENCODING=UTF-8 didn't work, so I can't say which config is causing this.

DukeValentine:
[...]

PR #551, opened 2019-11-01
Fix issue #548

DukeValentine:
I've taken care to minimally change the code so it doesn't affect what already works. 
At first I tried to just do a filename = filename.encode('utf-8'), but that turned out to cause breakage with the sqlite database. So I decided to just use a second filename, encoded in utf-8, for the the file related methods.


DukeValentine:
Sorry, I made a small mistake in one line. I'll retry the pull request

PR #571, opened 2019-11-16
Fix password check in getMemberInfoWhitecube

yzaoui:
I don't think `self._username` was meant to be checked twice, so this fixes that and prevents a `len(None)` situation.


Nandaka:
yep, my bad. copy paste error 😄

Issue #565, opened 2019-11-12
Fanbox Download by artist id issue - access denied


densetsumeru:
Using 20191111 updated version

### Description

Unable to use pixivutil to download from my supported fanbox creators

### Steps to Reproduce

Opened pixivdownloader
entered f2 download by artist id
entered artist ID
Max page = 0 or Max page = 1

**Expected behavior:** [What you expected to happen]

PixivUtil would download the various image posts for me.

**Actual behavior:** [What actually happened]

Following error text popped up:

Using 0Auth to retrieve member info for : [pixivID]
access_denied
Failed to login using 0Auth
press enter to exit.

pixivutil won't go any further.

### Versions
20191111

You can get this information from executing `PixivUtil2.py --help`.

yzaoui:
Been getting the same issue since I first started using PixivUtil2 two weeks ago. I've tried looking into the code to no avail. I feel like the culprit might be that `https://oauth.secure.pixiv.net/auth/token` has been moved? Can anyone else confirm that they don't get a 404 from requesting their token there?

Nandaka:
Weird, tested the PixivOAuth module and it able to login

yzaoui:
It specifically only happens with Fanbox downloads. Looking at it more closely I seem to be getting `{"has_error":true,"errors":{"system":{"message":"access_denied","code":1508}}}`. No proxies, username/password/cookie present in config.ini.

Searching the web, this error code came up recently when `X-Client-Time` and `X-Client-Hash` were added as requirements, but that's not a problem anymore. Any ideas what else it could be?

Syampuuh:
Also getting the issue, and I am also getting a 404 from `https://oauth.secure.pixiv.net/auth/token`.

Nandaka:
you will get 404 to that page if you supply it without any parameter. Check your PC clock because it quite time sensitive? If you have python 2.7 installed, maybe you can try to run it from sourcecode.

yzaoui:
Right, looks like it returns a 404 but it's not actually missing, that's just their response. But with the provided headers, it returns the 1508 error I mentioned above. I've been running it from sourcecode on python 2.7. I've synced up my time with the internet.

Nandaka:
@yzaoui get the latest commit and can you run the PixivOAuth.py in the console and copy the refresh token to `config.ini`. It should do self test to retrieve the artist info with oAuth header

[...]

PR #572, opened 2019-11-18
Fix LocalUTCOffsetTimezone string format for negative timezones

yzaoui:
For all timezones to the West of UTC, the current code formats it wrong. For instance, I live in the EST timezone, and the generated offset string is currently `--5:00` instead of the proper `-05:00`. The `X-Client-Time` header is thus always invalid, which is why #565 was happening to me.


Nandaka:
hahaha, that the reason why it failed login for the oAuth?

yzaoui:
Yup 😅 Ever since they started requiring `X-Client-Time`, fanbox never worked for anyone with negative timezones. I'm finally getting a refresh token for the first time ever, yay!

PR #603, opened 2019-12-18
Pylinted

NHOrus:
Well, I decided that this needs fixing. I think I hit most of tricky places.
Except exceptions, stuff can explode horribly on exceptions.


NHOrus:
I mean, now it's fully Py3, not the mix of Py3 and Py2 that works somehow.

PR #647, opened 2020-03-06
Refactor sanitize_filename in PixivHelper

alexwh:


Nandaka:
Possible to move out the re.compile out of the function? Considering this function being called multiple time, I think it is better to move it as properties, so it doesn't need to compile the regex everytime the function being called.

alexwh:
Ah, you're right, I missed that. I also removed the whitespace regex as they're handled later by the stripping of leading and trailing spaces or dots.

PixivHelper.py
```diff
@@ -59,94 +70,61 @@ def set_log_level(level):
...
     # Unescape '&amp;', '&lt;', and '&gt;'
-    s = html.unescape(s)
+    name = html.unescape(name)
 
-    # Replace badchars and badnames with _
-    name = __badchars__.sub('_', s)
-    if __badnames__.match(name.lower()):
+    # Replace badchars and reserved Windows file names with _
+    name = __badchars__.sub("_", name)
+    if Path(name).is_reserved():
```
Nandaka: this won't work if the name contains path separator, you still need to split based on the separator.
```
>>> Path("/aux/new").is_reserved() 
False
>>> os.makedirs("/aux/new")        
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python38-32\lib\os.py", line 211, in makedirs
    makedirs(head, exist_ok=exist_ok)
  File "C:\Python38-32\lib\os.py", line 221, in makedirs
    mkdir(name, mode)
NotADirectoryError: [WinError 267] The directory name is invalid: '/aux'
```

Nandaka: you can refer to https://github.com/Nandaka/PixivUtil2/blob/f1a1b926d2f0873e0d18f4cf45a897fbc6743211/PixivHelper.py#L92 for path split, maybe add the concatenation with '_' there.

Issue #656, opened 2020-03-29
Download fanbox articles in HTML


rwchunk:
### Prerequisites

* [x] Did you read FAQ section?
* [x] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?

### Description

I want to store and view the article's text with the pictures embedded.
Downloading article to an HTML file would preserve this.
The formatting is like a list of text lines and images.
I'm not really sure what would be the best way to go about this, but I'd really appreciate it.

bluerthanever:
Hey. I am actually quite interested in this idea.
I think technically it won't be toooooooo difficult. But I wanna ask a few questions.
1. How would you like the html pages to be like?  Exactly as what is shown on fanbox or a really simple html page with no style or whatsoever would do? 
2. What information would like to be shown on the page? Would you have the old txt files replaced by html files? Or do you want to control them in config.ini?

bluerthanever:
And 1 more, this should only apply to article type posts, right?

bluerthanever:
Hey. I am actually quite interested in this idea.
I think technically it won't be toooooooo difficult. Hmm.
I will probably start trying it out tomorrow on my side, and maybe pull a request?

bluerthanever:
#659

bluerthanever:
Hey. I just made it apply to article posts. is that enough for you?

bluerthanever:
Hey. I just made it apply to article posts, is that enough?
And firefox, chrome and edge seem to be working fine though, IE seems to be not so compatible. What browser do you use?

Nandaka:
PR merged

PR #659, opened 2020-04-01
Download fanbox articles in HTML #656

bluerthanever:
Added a switch "writeHtml" in config, to define whether to write fanbox article posts to html files in PixivConifg.
Added a property "linkToFile" for fanbox post class, to store each link and its local file path in a pairs as a dict, and a new method to write html files in PixivFanboxModel.
Added logic to store urls and local file paths when processing each image, and when the switch is on and the post is an article post, calls the method to write the html file.
Added a html pattern file for all the html files.


bluerthanever:
#659

bluerthanever:
I just modified it in my way...
And I tested on my side, works for me though, please confirm.

And if you see any other of my pull requests related to this, please ignore or delete those. I might have done something by mistake.

bluerthanever:
Made some more tests, works fine with firefox, chrome and edge on my side, but doesn't seem compatible with IE.... Probably need some help.

Nandaka:
is it possible to store the default html pattern to `PixivConstant.py` so if the `pattern.html` is deleted/missing, it will use the default pattern in the code?

bluerthanever:
I think that's not a problem.
So I will keep the old way, if the `pattern.html` file doesn't exist, it refers to `PixivConstant.py`, so if the user want to change something in the pattern, they can just do it themselves, right?

bluerthanever:
Hmm... Just saw the new release of beta version. Maybe I should have used "template" instead of "pattern"....

Nandaka:
I also changed the file detection logic a bit:
https://github.com/Nandaka/PixivUtil2/blob/master/PixivUtil2.py#L2070-L2076

bluerthanever:
Yes. I have noticed.

PR #660, opened 2020-04-02
print_progress fix & refactor

wmjdgla:
- Combined printing of '\r' and progress string into single print statement
- Changed .format(...) to f-string expressions
- Changed hardcoded progress bar length to variable (animBarLen)
- Fixed jerky progress bar for indeterminate case. The progress bar now scrolls smoothly. The original progress bar starts with
`[|||....................]`
and ends with
`[...................|||.]`
and then jumps back to start.
This can be shown by the following test with the original code:
```
# function hollowed out to simplify test code
def size_in_str(totalSize):
    return totalSize

import time
curr = 0
while True:
    pos = curr % 20
    anim = "{0}{1}".format('.' * pos, '|||' + '.' * (20 - pos))
    print('\r', end=' ')
    print('[{0:20}] {1}'.format(anim, size_in_str(curr)), end=' ')
    curr += 1
    time.sleep(0.3)
```

### DIFF
--- a/PixivHelper.py
+++ b/PixivHelper.py
@@ -642,19 +642,20 @@ def download_image(url, filename, res, file_size, overwrite):
 def print_progress(curr, total):
     # [12345678901234567890]
     # [||||||||------------]
-
+    animBarLen = 20
+    
     if total > 0:
-        complete = int((curr * 20) / total)
-        print('\r', end=' ')
-        msg = '[{0:20}] {1} of {2}'.format('|' * complete, size_in_str(curr), size_in_str(total))
-        print('{0}'.format(msg), end=' ')
+        complete = int((curr * animBarLen) / total)
+        print(f"[{'|' * complete:{animBarLen}}] {size_in_str(curr)} of {size_in_str(total)}", end='\r')
     else:
         # indeterminite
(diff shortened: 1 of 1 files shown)


bluerthanever:
Hey.  After this modification, there's some thing funny showing when I launch from source code in cmd.
When downloading:
QQ截图20200404233041

I think this is because this line here below...
No idea what it means. It just prints "                       0 bytes" on my side.
https://github.com/Nandaka/PixivUtil2/blob/1bc227b4666bcf0f7737daa1c1147e8fd69e1de8/PixivHelper.py#L593


Download completed:
QQ截图20200404233242
Before, it's like this
QQ截图20200404234211
I think this is because it changed the location of "\r". 
Before then "\r" was printed first when print_progress was called and ends with a " ", and when the file is downloaded, the "completes in ..." was printed after that.
But now "\r" is printed last, and when the file is downloaded, which means the last "print_progress" was called, the cursor was moved to the beginning, and then the whole "completes ...." sentence is printed from the start, which is shorter than the progress bar.

wmjdgla:
Ah you're right, my bad. Will make the correction in a new PR. Thanks!

________________________________

PR #661, opened 2020-04-04
Added logic for links if links token was in p blocks

bluerthanever:
Sometimes p blocks contain links. Added logic for that.


bluerthanever:
I think this is not too complicated so I didn't test this.
Logically I think it's correct.  But I don't see posts with more than 1 links in p blocks, so I can't really say I am 100% sure.... Let's just fix it when it happens.

Nandaka:
I didn't catch it, do you have a sample post with link inside the p tag?

bluerthanever:
[sample.json.log](https://github.com/Nandaka/PixivUtil2/files/4431077/sample.json.log)
Please refer to the attachment.

Nandaka:
if possible, can you add an unit test in test_PixivModel_fanbox.py using that json (maybe with a case with multiple link in single p tag.

bluerthanever:
Hmm. I have not encountered any post with multiple links in a single p tag yet. I can only make one, which I can not make sure if it's going to match how pixiv does it....
I will try to add a unit test.

bluerthanever:
Good thing you had me add the tests... haha.
Btw, after the modification, the body_text is longer than before,, thub, some numbers in other tests doesn't pass anymore.

bluerthanever:
Just saw the beta.... was thinking maybe I should have used "template" instead of "pattern"....

Nandaka:
can you update it for me?

bluerthanever:
Haha. I guess so. And some other modifications later.

bluerthanever:
Done.
BTW, if I reply to closed pull requests, can you get notifications? I just replied to #660

PR #664, opened 2020-04-05
Progress bar print fix

bluerthanever:
https://github.com/Nandaka/PixivUtil2/pull/660#issuecomment-609050617

Hey.  After #660 , there's some thing funny showing when I launch from source code in cmd.
When downloading:

Download completed:

Before, it's like this

I think this whole funny print(?) is because the location of "\r" is changed from end to beginning. 
Before then "\r" was printed first when print_progress was called and ends with a " ", and when the file is downloaded, the "completes in ..." was printed after that.
But now "\r" is printed last, and when the file is downloaded, which means the last "print_progress" was called, the cursor was moved to the beginning, and then the whole "completes ...." sentence is printed from the start, which is shorter than the progress bar.

And for the first image, is also because of the code in the link below, no idea what it means. It just prints "                       0 bytes" on my side. Maybe it was meant to be overwritten by the "\r" and progress bar? Like a place holder?
https://github.com/Nandaka/PixivUtil2/blob/1bc227b4666bcf0f7737daa1c1147e8fd69e1de8/PixivHelper.py#L593

### DIFF
--- a/PixivHelper.py
+++ b/PixivHelper.py
@@ -644,10 +644,10 @@ def print_progress(curr, total):
     # [12345678901234567890]
     # [||||||||------------]
     animBarLen = 20
-
+    print('\r', end=' ')
     if total > 0:
         complete = int((curr * animBarLen) / total)
-        print(f"[{'|' * complete:{animBarLen}}] {size_in_str(curr)} of {size_in_str(total)}", end='\r')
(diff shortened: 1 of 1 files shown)


Nandaka:
woops, I already made changes to fix this.

bluerthanever:
Haha. That's cool.

PR #671, opened 2020-04-17
Html improvements

bluerthanever:
1. Users can now decide whether to use absolute paths or relative paths in html with useAbsolutePathsInHtml in config.ini. When set to False, util uses relative paths.
2. Renamed some parameters for related methods and constant names, and filename of user-definable template.
3. Added logic to use relative paths in html.
4. Added right padding for p tags in html style.
I tested on my side, and it's working fine.

### DIFF
--- a/PixivConfig.py
+++ b/PixivConfig.py
@@ -51,6 +51,7 @@ class PixivConfig():
     writeImageInfo = False
     writeImageJSON = False
     writeHtml = False
+    useAbsolutePathsInHtml = False
     verifyImage = False
     writeUrlInDescription = False
     urlBlacklistRegex = ""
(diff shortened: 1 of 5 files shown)


bluerthanever:
4. Added right padding for p tags in html style.
I tested on my side, and it's working fine.

PixivConfig.py
```diff
@@ -637,6 +638,13 @@ def loadConfig(self, path=None):
                 print("writeHtml = False")
                 haveError = True
 
+            try:
+                self.useAbsolutePathsInHtml = config.getboolean('Settings', 'useAbsolutePathsInHtml')
+            except ValueError:
+                self.useAbsolutePathsInHtml = False
```
Nandaka, line 644: change the default to True? because the old behaviour is always using Absolute Path right?

bluerthanever: I hesitated on that as well, but eventually I decided to set the default False. 

Because firstly, I think relative paths is more "HTML". For example when you save a page from browser, it uses relative paths. 
And secondly, there's no logic to detect whether the file exists or not, so if the old file exists, it will be overwritten no matter what. And there could only be two occasions when user uses the util, 1, user didn't move the files at all, which I means means the at a big chance that the user don't need to move the files, so the change from absolute paths won't effect anything, the pages would work just fine; 2. no html files were in the same folder, which means user moved it elsewhere or deleted it, so the new file would still work. But when they move it again, they will find out that it doesn't work so they should change the configuration.

What I meant was that for users who don't move their files or change the structures of them, setting the default to False, would save an step to change the settings. Which I think most users would not move the html files solely anywhere around.....

Sorry for the poor explanation. Not sure I made it clear.

PR #676, opened 2020-04-25
Simplify(?) codes with reflection(?) thoughts

bluerthanever:
1) Created a new class `ConfigItem`
2) Created a private(?) variable `__items` inside PixivConfig, which is a list, to store all config items
3) In each method of `PixivConfig`, iterate `__items` to get all config values
4) When there's error in `loadConfig`, uses self.configFileLocation as path parameter explicitly for `writeConfig`

Notes:
1) Default values are kind of a mixture of preset and values inside `loadConfig`
2) Because of the iteration, `printConfig` output is not in the same order as in the old way
3) I tested on my side, between the old way and the new way, time elapsed of each method doesn't differ much, or at least, that the difference doesn't matter much...

### DIFF
--- a/PixivConfig.py
+++ b/PixivConfig.py
@@ -248,6 +248,5 @@ class PixivConfig():
         for k, g in groups:
             print(f" [{k}]")
             for item in g:
-                print(f" - {item.option:.{30}} = {self.__getattribute__(item.option)}")
-
+                print(f" - {item.option:{25}} = {self.__getattribute__(item.option)}")
         print('')
--- a/PixivConfig.py
+++ b/PixivConfig.py
@@ -152,6 +152,7 @@ class PixivConfig():
     def __init__(self):
         for item in self.__items:
             setattr(self, item.option, item.default)
+        self.proxy = {'http': self.proxyAddress, 'https': self.proxyAddress}
 
     def loadConfig(self, path=None):
         if path is not None:
@@ -243,10 +244,14 @@ class PixivConfig():
 
     def printConfig(self):
         print('Configuration: ')
-
-        groups = itertools.groupby(PixivConfig.__items, lambda x: x.section)
-        for k, g in groups:
-            print(f" [{k}]")
-            for item in g:
-                print(f" - {item.option:{25}} = {self.__getattribute__(item.option)}")
+        groups = {k: list(g) for k, g in itertools.groupby(PixivConfig.__items, lambda x: x.section)}
+        sections = ["Authentication", "Network", "Debug", "IrfanView", "Settings", "Filename", "Pixiv", "FFmpeg",
+                    "Ugoira", "DownloadControl"]
+        sections.extend([k for k in groups if k not in sections])
+        for section in sections:
+            g = groups.get(section)
+            if g:
+                print(f" [{section}]")
+                for item in g:
+                    print(f" - {item.option:{25}} = {self.__getattribute__(item.option)}")
         print('')
--- a/PixivConfig.py
+++ b/PixivConfig.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import configparser
+import itertools
 import os
 import os.path
 import shutil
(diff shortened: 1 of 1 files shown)


Nandaka:
how to consider the order of the item? 
There is a reason that you need to include the new configItem in the last when you read the config file, so when it recreate the new config,  it can keep the old values to include both old and new item.

bluerthanever:
Hmm, if your main concern is the order of each item and old values from old `config.ini` file,, I ran some test. While doing that I found out I forgot to deal with when file doesn't exist.

Basically I just delete some items or alter the value of some items, or change the orders of some sections, then load them and print/write them to check the results:
1. For `printConfig` and `writeConfig`, the outputs are in the same order as that written in `__items` in `PixivConfig`
2. If some values given in the file are illegal, other legal values would be kept. For example, if there is no username option in the file, all other values would be saved, while it would be reset if it's the old way.

Over all,  I think the new codes would make sure that the legal values in the old file would be kept.
But there are 2 main difference:
1. The output of `printConfig` would be different from the the old way.
2. When config file doesn't exist, instead of one single line output, all items and default values would be printed on the screen.

Please confirm.

bluerthanever:
And I think there would be 2 main benefits:
1. Codes would be neater. Not sure if that's the word to use but I think you get the idea, haha
2. Codes would be easier to maintain. You just need to change one place, no matter if it's that you want to add new sections or items, or change the whole structure

bluerthanever:
Or I could override the `__init__` to set all items to default values first?

bluerthanever:
Sorry for multiple submission. I just submit the final version... This eliminate the second difference mentioned above.

bluerthanever:
I forgot to mention that when it comes to `mangaFilenameFormat`, instead of appending `%page_index%` to its end when it doesn't contain `%urlFilename%`, `%page_index%` or `%page_number%`, the format would be considered invalid and would be replaced with default value.

PixivConfig.py
```diff
@@ -11,648 +12,192 @@
...
-                self.processFromDb = True
-                haveError = True
-
-            try:
-                self.proxyAddress = config.get('Network', 'proxyaddress')
-            except ValueError:
-                print("proxyAddress = ''")
-                self.proxyAddress = ''
-                haveError = True
-            self.proxy = {'http': self.proxyAddress, 'https': self.proxyAddress}
```
Nandaka, line 161: missed out this line for followup call?

bluerthanever: Hmm.... you mean `self.proxy` should be a followup of `proxyAddress`?

Nandaka: ah, see it in: https://github.com/Nandaka/PixivUtil2/blob/b97865208fdf9cacc6855d2186ece453a715cc23/PixivConfig.py#L196

Nandaka:
For printing, I think you can sort by the section first and then followed by the config item.

bluerthanever:
Hmm... Do you mean to print the sections in the same order as before? 
It prints the section name followed by the items in it, then another section and so now already...

PixivConfig.py
Nandaka: I think you can keep the current implementation to keep it simple first, else you need to modify the followup lambda to pass the caller, so it can set the attributes in the caller.

bluerthanever: Actually this could be done with `followup` as well and it should be better.
Nice point. Haha

Nandaka:
Ya, you do groupby the section name, is this section name is sorted, or based on the order you add (so in this case Network will show up first).

PixivConfig.py
bluerthanever: okay... but I think I maybe need to set `self.proxy` in `__init__` as well, do you think so?

bluerthanever:
Okay. I will work on it now.

bluerthanever:
Done. Please confirm. And for the `self.proxy`, I think it would be kinda hard to be implemented with the current followup, whatever lambda it passes. But I added one `self.proxy = {"http": self.proxyAddress, "https": self.proxyAddress, }` in `__init__`, not crucial but I guess it's better?

PR #680, opened 2020-04-27
FANBOX download enhancement

bluerthanever:
1. Some minor format changes made by pyCharm
2. Moved codes to download FANBOX cover images from `processFanboxArtist` to `processFanboxImages`, because I don't think cover images should be downloaded if the post is not accessible
3. Post infomation (mainly just artist_id, post_id, title, feeRequired, workDate, updatedDate, type) would be written into database first thing after 'processFanboxImages' is called inside it, before retrieving post data row from database for comparison between the newly retrieved post updated date and that written in the database. Post would not be processed if it is restricted or download history exists. 
4. Then after everything is down, images downloaded and image info written to file, update the `updatedDate` in database
5. Added entry for download fanbox post by post ids, and some related logic in option parser thiingy

### DIFF
--- a/PixivBrowserFactory.py
+++ b/PixivBrowserFactory.py
@@ -313,7 +313,7 @@ class PixivBrowser(mechanize.Browser):
             login_cookie = self._config.cookieFanbox
 
         if len(login_cookie) > 0:
-            PixivHelper.print_and_log('info', 'Trying to log in with saved cookie')
+            PixivHelper.print_and_log('info', 'Trying to log in FANBOX with saved cookie')
             self.clearCookie()
             self._loadCookie(login_cookie, "fanbox.cc")
 
@@ -321,10 +321,15 @@ class PixivBrowser(mechanize.Browser):
             req.add_header('Accept', 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8')
             req.add_header('Origin', 'https://www.fanbox.cc')
             req.add_header('User-Agent', self._config.useragent)
-            res = self.open_with_retry(req)
-            parsed = BeautifulSoup(res, features="html5lib").decode('utf-8')
-            PixivHelper.get_logger().info('Logging in, return url: %s', res.geturl())
-            res.close()
+            try:
+                res = self.open_with_retry(req)
+                parsed = BeautifulSoup(res, features="html5lib").decode('utf-8')
+                PixivHelper.get_logger().info('Logging in, return url: %s', res.geturl())
+                res.close()
+            except BaseException as e:
+                PixivHelper.get_logger().error('Error at fanboxLoginUsingCookie, please check cookieFanbox: '
+                                               + format(sys.exc_info()))
+                return False
 
             result = False
             if '"user":{"isLoggedIn":true' in str(parsed):
--- a/PixivBrowserFactory.py
+++ b/PixivBrowserFactory.py
@@ -725,14 +725,30 @@ class PixivBrowser(mechanize.Browser):
                     PixivHelper.toUnicode(response)))
 
     def fanboxGetSupportedUsers(self):
-        ''' get all supported users from the list from https://fanbox.pixiv.net/api/plan.listSupporting'''
-        url = 'https://fanbox.pixiv.net/api/plan.listSupporting'
+        url = 'https://api.fanbox.cc/plan.listSupporting'
         PixivHelper.print_and_log('info', f'Getting supported artists from {url}')
-        referer = "https://www.pixiv.net/fanbox/support/creators"
+        referer = "https://www.fanbox.cc/creators/supporting"
         req = mechanize.Request(url)
         req.add_header('Accept', 'application/json, text/plain, */*')
         req.add_header('Referer', referer)
-        req.add_header('Origin', 'https://www.pixiv.net')
+        req.add_header('Origin', 'https://www.fanbox.cc')
+        req.add_header('User-Agent', self._config.useragent)
+
+        res = self.open_with_retry(req)
+        # read the json response
+        response = res.read()
+        res.close()
+        result = Fanbox(response)
+        return result
+
+    def fanboxGetFollowedUsers(self):
+        url = 'https://api.fanbox.cc/creator.listFollowing'
+        PixivHelper.print_and_log('info', f'Getting supported artists from {url}')
+        referer = "https://www.fanbox.cc/creators/following"
+        req = mechanize.Request(url)
+        req.add_header('Accept', 'application/json, text/plain, */*')
+        req.add_header('Referer', referer)
+        req.add_header('Origin', 'https://www.fanbox.cc')
         req.add_header('User-Agent', self._config.useragent)
 
         res = self.open_with_retry(req)
@@ -778,21 +794,6 @@ class PixivBrowser(mechanize.Browser):
 
         for post in result.posts:
             js = self.fanboxGetPost(post.imageId, artist_id)
-                                                                     
-                                                                                   
-                                                                                               
-                                                                                
-                                            
-                                                                           
-                                                  
-                                                               
-                                                                  
-
-                                               
-                                     
-                                                                     
-                         
-                                           
             post.parsePost(js["body"])
 
         return result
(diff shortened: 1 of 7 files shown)


bluerthanever:
Wish I didn't miss anything....

bluerthanever:
...I am confused now about which I should be using, `following` or `followed`...

I wanted to change the apis, I mean I think it's the best, even though the old ones are still good to use.... But some classes might need to be changed a lot... and I am like tired now and don't wanna think about it anymore.... Changed some simple ones though.

bluerthanever:
Hmm... I found a problem that is to use the new url (https://api.fanbox.cc/...), there need to be a new cookie..... urr.

bluerthanever:
I will just add a new item in config, and users would need to copy and paste the new `FANBOXSEESID` into config in order to use it.... how is that....

bluerthanever:
Okay.... I guess I am almost done.... I will submit the changes now..

bluerthanever:
Changes required... that I can think of...
1. Users need to add the new `cookieFanbox` which is `FANBOXSESSID` when visiting fanbox.cc.
2. This cookie is not obtainable by codes... or not implemented currently

bluerthanever:
The `FANBOXSESSID` seems to be changing constantly....hmmm

Nandaka:
any changes on the unit test? ensure the existing unit test doesn't break, else I need you to update it (as I don't have fanbox access).

bluerthanever:
hmmm... I hate unit test....
I ran it and downloaded stuff...haha. I will take a rest, and take a look at unit tests later

Nandaka:
yeah, you change the structure and it fail to import the class
image

bluerthanever:
Yeah. Haha. Later tonight.

Nandaka:
tried to fix the unittest, but looks like the source json also changed?
I need to set the unittest to pass, else I cannot commit back as I have enable Travis CI 
image

bluerthanever:
Yeah... I noticed some minor changes, like some new tokens added. But mostly the same...
So I will just fix the syntax to make it pass now...


What is that interface though...?

Nandaka:
That one is from TravisCI, so if the build pass then I can merge the pull request, like below:
image

Nandaka:
merged 😄

bluerthanever:
Hmm.. Didn't notice that before... cuz it was always green before...
Actually I think I can try to do login with account ans password...? do you think that's necessary? I just caught some packages today.... was thinking about how to obtain fanbox cookie then....

Nandaka:
> account and password

currently not possible due to captcha blocking the authentication form, right?

bluerthanever:
Actually I am not sure that's like impossible.... cuz I don't see any captcha authentication form on the page, and if it's hidden I think it could be bypassed or simulated with with `execjs` if it's in js or similar logic if not...?

bluerthanever:
well I tried to do login with twitter though...
anyway I don't think this should be prioritized.... and I might need further studies on that....

Nandaka:
hmm, I'm not sure, I think it is depend on your login behaviour and your IP address? I also don't use twitter or facebook anymore.

bluerthanever:
I'm not sure either... 
the important thing is that it's usable, then comes convenience.
I will look into this if I am like really bored or something. haha

AgentThirteen:
Thanks for all your work so far bluerthanever.
I was about to write an issue for this but it looks like it was an intended enhancement.

> Moved codes to download FANBOX cover images from processFanboxArtist to processFanboxImages, because I don't think cover images should be downloaded if the post is not accessible

FANBOX cover images can be used if you pause then resume support to see if anything has changed as well as keep track of deleted posts. Any intent to bring them back as an option if the coding isn't too much of a hassle? It doesn't really slow down the process all that much (at least not until #686 - I will try to keep that one issue informed as I was short on time but should be able to test with a VPN or proxy at some point).

Users who have new creator post notifications set can have the cover sent to their email address but that is a rather buggy thumbnail and not a default setting if I am not mistaken.

bluerthanever:
@AgentThirteen 
Well, it was not on the speed side, but more on the file management side when I was moving the codes, because sometimes I download from an creator with free posts and paid ones, and there would be a whole lot of single cover images.
And in this enhancement, I recorded the `updatedDate` of each post which is retrieved from FANBOX API in database, so if anything has changed, I think the date is changed too and when the post is hit again it will be downloaded again.

Unless you would want those cover images intentionally?

And I am curious on how you keep track of deleted posts with cover images, do you mind sharing the idea? Not saying no yet, but curious about how different users use it.

AgentThirteen:
Thank you for the quick reply.

Oh I see. That certainly makes sense if you have a lot of followed creators considering how many have a FANBOX by now. I kind of missed the updatedDate enhancement, sorry, this is good to know for the pause-and-resume usage.

As for keeping track of deleted posts with cover images, simply checking based on notifications and locally updated date as you stated as I don't use the API myself and I am not really a coding guru let alone an amateur, haha.

bluerthanever:
> As for keeping track of deleted posts with cover images, simply checking based on notifications and locally updated date as you stated as I don't use the API myself and I am not really a coding guru let alone an amateur, haha.

Sorry I don't quite get this. Do you mean you keep all notification emails of an artist's new post and compare to find out which posts are deleted? Because I don't remember that FANBOX would notify users if an artist deletes their posts.

But whichever way, if you still need to download the cover image, please don't hesitate to let me know.

AgentThirteen:
Yes, that's exactly what I meant. I actually keep all mail sent as updated/new post notifications as I have pretty much unlimited email storage space and it's easy to parse regardless of the amount received. A thumbnail suddenly going missing indicates that the post is gone (the same thing will happen in bell notifications but that kind of history isn't fully checkable as far as I could tell) but there are surely easier ways to tell I am not aware of.

> But whichever way, if you still need to download the cover image, please don't hesitate to let me know.

That'd be much appreciated if that's possible when you have time, thanks a lot.

Not sure how many users need this though, so probably as an optional feature if you feel like that's a lot of clogging?

bluerthanever:
@Nandaka Hey, Nandaka-san. I would like to add the following enhancements, what do you think?
1. File formats for FANBOX posts
 - 3 separate formats for cover images, images inside posts, and post info
2. Options to write HTML for posts that are not article type
 - an option named `whenTextsLongerThan` or something to write posts into HTML, when they contains more than 1 image, and texts longer than that number
 - a new HTML template for such posts (that are not articles) or modify the current one so it could be used for both
3. An option to decide whether to download cover images when post is restricted
4. An option to decide whether to check database download history to decide whether to download a post or not
5. Move FANBOX options together under `FANBOX` section

Nandaka:
1. You means the filename formats? Currently it is already using `filenameFormat` for the cover, `filenameMangaFormat` for the images inside the post, and `filenameInfoFormat` for the post info. Any reason to create separate filename formats?
2. Should be ok.
3. Should be ok.
4. Is it something like this? Only download the post if updatedDate is different?
```
if config.checkFanboxPostHistory:
   result = db.getFanboxPostHistory(post_id)
   if result is not None:
      post_data = br.getFanboxPost(post_id)
      if result.updatedDate == post_data.updatedDate:
         return False # skip the post
   return True # download the post
```
5. Sure, ensure the config upgrate can be handled. I think these are the existing config for Fanbox?
   - `writeHtml`
   - `useAbsolutePathsInHtml`

bluerthanever:
1. Well, it seems like some users are confused, to make it easier to understand, and also, for users who would want the FANBOX posts to be saved by different formats and are not willing to use a different config. Haha, for the lazy ones?
4. Yup, something like that.
5. Hmm, Now that you've mentioned it, and I just realized that I am moving the sections, which the current structure does not support yet. I will find a way to deal with that.

Nandaka:
well, ok. For moving section upgrade, no need to think on the way to do it automatically, just ensure all the previous configs are read then only throw the error message.

On the previous structure, I'll just move the reading code to the last line after I update the section name (treat it like new config).

bluerthanever:
Well currently, if a config is moved to another section, it will be reset... so not technically ensuring all previous configs are read... So I need to think of a solution I guess.

PR #683, opened 2020-04-28
Fixing a mistake when dealing with downloading by post id

bluerthanever:
Sorry I missed a spot...
This is new and not in the unit test....
Sorry

### DIFF
--- a/PixivBrowserFactory.py
+++ b/PixivBrowserFactory.py
@@ -878,7 +878,7 @@ class PixivBrowser(mechanize.Browser):
         # https://www.pixiv.net/fanbox/creator/104409/post/279561
         p_url = f"https://api.fanbox.cc/post.info?postId={post_id}"
         # referer doesn't seeem to be essential
-        p_referer = f"https://www.fanbox.cc/@{artist.creatorId}/posts/{post_id}"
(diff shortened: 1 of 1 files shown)


Nandaka:
can you add a new unittest so it can catch the new usecase? 
Thanks a lot for the fanbox development and test, because I don't use it at all 😅

bluerthanever:
Haha. This is killing me... I hate unit tests.
Should I add it here or pull a new request? And is there a deadline...?

Nandaka:
nah, just use the same PR for the unit test.
You need to add unit test, else your future self will kill you 😆 , also good for testing regression.

bluerthanever:
haha.. ok.... I will try to finish it tonight.

Nandaka:
take it easy, it won't kill you to push it later.

bluerthanever:
That's fine... but can you add unit tests that requires internet?
cuz the new function kinda requires that, otherwise I think it's already covered by other tests, if it's just reading jsons and parsing them into instances...
The mistake in this PR happened when building requests

Nandaka:
hmm, I don't think Travis can connect to the internet for the free version, also that require you to pass the authentication data.

bluerthanever:
Okay... I guess I was always a little sloppy... 
I will try not to make mistakes next time. 😄

PR #732, opened 2020-06-27
Download progress print, config, and fanbox artists

bluerthanever:
1. Some output print optimization for downloading
    before:
 ....|||.............] 805.77 KiB      ] 805.77 KiB
 Completed in 1.223042s (658.82 KiB/s)
 done.
    after:
[....|||.............] 805.77 KiB  Completed in 1.223042s (658.82 KiB/s)
 done.

2. Load config before processing each FANBOX aritst in menu_fanbox_download_from_artist
3. Some changes for PixivConfig to avoid writing configuration to file if it's not changed
- 1. Added class field `__props` in `PixivConfig` to store all option names
- 2. Added instance field '__initialized` and `__modified`
- The above will be used in `__setattr__`
- `__initialized` will be set to true at the end of `__init__`
- 
- 1. If the instance is initialized, and the new attribute is in `__props` and the attribute's value is new, `__modified` would be set to true with overriden method `__setattr__`
- 2. When calling `writeConfig`, first check `__modified`, if it's true, then continue and set `__modified` to false, otherwise return

### DIFF
--- a/PixivDownloadHandler.py
+++ b/PixivDownloadHandler.py
@@ -246,7 +246,7 @@ def perform_download(url, file_size, filename, overwrite, config, referer=None, 
     if referer is None:
         referer = config.referer
     # actual download
-    PixivHelper.print_and_log(None, '\rStart downloading...', newline=False)
+    PixivHelper.print_and_log(None, 'Start downloading...\r', newline=False)
     # fetch filesize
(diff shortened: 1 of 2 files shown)


bluerthanever:
Sorry, wrong branch

PR #733, opened 2020-06-27
Download progress print, config, and fanbox artists

bluerthanever:
1. Some output print optimization for downloading
    before:
 ....|||.............] 805.77 KiB      ] 805.77 KiB
 Completed in 1.223042s (658.82 KiB/s)
 done.
    after:
[....|||.............] 805.77 KiB  Completed in 1.223042s (658.82 KiB/s)
 done.

2. Load config before processing each FANBOX aritst in menu_fanbox_download_from_artist
3. Some changes for PixivConfig to avoid writing configuration to file if it's not changed
- 1. Added class field `__props` in `PixivConfig` to store all option names
- 2. Added instance field '__initialized` and `__modified`
- The above will be used in `__setattr__`
- `__initialized` will be set to true at the end of `__init__`
- 
- 1. If the instance is initialized, and the new attribute is in `__props` and the attribute's value is new, `__modified` would be set to true with overriden method `__setattr__`
- 2. When calling `writeConfig`, first check `__modified`, if it's true, then continue and set `__modified` to false, otherwise return

### DIFF
--- a/PixivUtil2.py
+++ b/PixivUtil2.py
@@ -1023,9 +1023,9 @@ def menu_fanbox_download_from_artist(op_is_valid, via, args):
         PixivHelper.print_and_log("info", f"No {via_type} artist!")
         return
     PixivHelper.print_and_log("info", f"Found {len(artists)} {via_type} artist(s)")
-    print(", ".join(str(artists)))
 
     for artist in artists:
+        __config__.loadConfig(path=configfile)
         __br__.fanboxUpdateArtistToken(artist)
         # Issue #567
         try:
--- a/PixivConfig.py
+++ b/PixivConfig.py
@@ -41,6 +41,8 @@ class PixivConfig():
     '''Configuration class'''
     __logger = PixivHelper.get_logger()
     configFileLocation = "config.ini"
+    __initialized = False
+    __modified = False
 
     __items = [
         ConfigItem("Network", "useProxy", False),
@@ -166,10 +168,23 @@ class PixivConfig():
 
     proxy = {"http": "", "https": "", }
 
+    __props = [x.option for x in __items]
+
     def __init__(self):
         for item in self.__items:
             setattr(self, item.option, item.default)
         self.proxy = {'http': self.proxyAddress, 'https': self.proxyAddress}
+        self.__initialized = True
+        self.__modified = False
+
+    def __setattr__(self, key, value):
+        if self.__initialized:
+            if key in PixivConfig.__props:
+                if self.__getattribute__(key) != value:
+                    self.__modified = True
+                else:
+                    return
+        super(PixivConfig, self).__setattr__(key, value)
 
     def loadConfig(self, path=None):
         if path is not None:
@@ -227,11 +242,13 @@ class PixivConfig():
         if haveError:
             print('Configurations with invalid value are set to default value.')
             self.writeConfig(error=True, path=self.configFileLocation)
-
+        self.__modified = False
         print('done.')
 
     # -UI01B------write config
     def writeConfig(self, error=False, path=None):
+        if not self.__modified:
+            return
         '''Backup old config if exist and write updated config.ini'''
         print('Writing config file...', end=' ')
         config = configparser.RawConfigParser()
@@ -267,7 +284,7 @@ class PixivConfig():
         except BaseException:
             self.__logger.exception('Error at writeConfig()')
             raise
-
+        self.__modified = False
         print('done.')
 
     def printConfig(self):
--- a/PixivHelper.py
+++ b/PixivHelper.py
@@ -629,13 +629,11 @@ def download_image(url, filename, res, file_size, overwrite):
             # check if downloaded file is complete
             if file_size > 0 and curr == file_size:
                 total_time = (datetime.now() - start_time).total_seconds()
-                print_and_log(None, "")
                 print_and_log(None, f' Completed in {total_time}s ({speed_in_str(file_size, total_time)})')
                 break
 
             elif curr == prev:  # no file size info
                 total_time = (datetime.now() - start_time).total_seconds()
-                print_and_log(None, "")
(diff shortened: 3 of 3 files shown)


PixivUtil2.py
```diff
@@ -1023,9 +1023,9 @@ def menu_fanbox_download_from_artist(op_is_valid, via, args):
         PixivHelper.print_and_log("info", f"No {via_type} artist!")
         return
     PixivHelper.print_and_log("info", f"Found {len(artists)} {via_type} artist(s)")
-    print(", ".join(str(artists)))
 
     for artist in artists:
+        __config__.loadConfig(path=configfile)
         __br__.fanboxUpdateArtistToken(artist)
```
Nandaka, line 1029: any reason not to move it inside processFanboxArtist()? it is called in menu_fanbox_download_from_artist() and menu_fanbox_download_by_artist_or_creator_id, so at least the behaviour is consistent for those menu.

bluerthanever: well, briefly speaking, it's just simpler.
specifically speaking:
1. the first parameter for `processFanboxArtist` is a `FanboxAritst` class item
2. another menu to download fanbox with id also calls `processFanboxArtist` after it gets the `FanboxArtist` instance with browser `fanboxGetArtistById` inside which calls the `fanboxUpdateArtistToken` method
3. currently when downloading from supporting list or following list, the menu method first gets a list of `FanboxArtist` instance with `fanboxGetUsers` method, and then passes each artist instance to `processFanboxArtist`, in which it calls the `fanboxUpdateArtistToken` to update some other fields of the instance before proceeding with the posts.

So if i need to put config.loadConfig inside `fanboxUpdateArtistToken`, I need to:
1. make `Fanbox.parseArtists` return a list of ids
2. change the first parameter of `processFanboxArtist` to an id instead of `FanboxAritst` instance
3. put `config.loadConfig` inside `fanboxUpdateArtistToken`
4. put `artist = __br__.fanboxGetArtistById` inside `fanboxUpdateArtistToken` after 3.
5. delete calling of `__br__.fanboxGetArtistById` and `__br__.fanboxUpdateArtistToken` in related menu methods
6. preferably change the name for some methods to make the name clear and match its content

or there could be some other ways as well.... but it's not that it's difficult, just too much fuss....

and I from the top am against that the config should reload between artists. it's like not a regular operation.
it's like being given a list of jobs, and instead of having the same criteria, or a set of criteria for each at the start, the giver calls in and checks on you and changes the criteria in between the jobs. or that a program is running and someone wants to cut in and change the value of one parameter.
this kind of case exists and makes sense but it's not how it should be.
i think the users would always have all different kinds of ideas, sometimes absurd. maybe next time he would say, the process of one creator is too fast, please insert a prompt in between the creators so i have time to change the configuration. sometimes mine are absurd too. 

anyway, i am just venting. if you would like to have them changed to the later form, please let me know.

Nandaka:
nah, if that the case, then I think only `__config__.loadConfig(path=configfile)` is the one needed to be moved inside `processFanboxArtist()`

bluerthanever:
well, if **ONLY** `__config__.loadConfig(path=configfile)` is moved inside `processFanboxArtist()`, then for `menu_fanbox_dowload_by_artist_or_creator_id`, the sequence would be:
1. creating a `FanboxArtist` instance with `__br__.fanboxGetCArtistById`, in which `__config__.writeConfig` would be called
2. go into `processFanboxArtist`, and `loadConfig` will be called which i don't think is necessary

and in `menu_fanbox_download_from_artist`, for each `FanboxArtist` instance:
1. ` __br__.fanboxUpdateArtistToken` is called firstly, in which `__config__.writeConfig` would be called
2.  go into `processFanboxArtist`, and `loadConfig` will be called 
this order seems weird to me.... unless you are talking about putting '__config__.loadConfig(path=configfile)` at the end

Nandaka:
I see, because of the side effect of `__br__.fanboxUpdateArtistToken` causing the confgi to be updated, but there is a risk of overwriting if the config was changed in between.
If that the case, then maybe adding `__config__.loadConfig(path)` before calling `processFanboxArtist()` in ' menu_fanbox_download_by_artist_or_creator_id' will do?

bluerthanever:
huh? you mean still move `__config__.loadConfig(path=configfile)` inside `processFanboxArtist`, and then add `__config__.loadConfig(path)` before calling `processFanboxArtist` in `menu_fanbox_download_by_artist_or_creator_id`?
so when processing a single id, the order would be `loadConfig`, then `fanboxGetArtistById` in which `writeConfig` will be called, then enters `processArtist`, and `loadConfig` again? **load -> write -> load**
and for supporting or following artists, it's **write -> load**?

bluerthanever:
Never mind. I just found another place to optimize, I will close this one and change it to the later form.

Nandaka:
No need to make it so complicated.
```
def menu_fanbox_download_by_artist_or_creator_id(op_is_valid, args):
    __log__.info('Download FANBOX by Artist or Creator ID mode.')
    end_page = 0
    id = ''

    if op_is_valid and len(args) > 0:
        id = args[0]
        if len(args) > 1:
            end_page = args[1]
    else:
        id = input("Artist/Creator ID = ").rstrip("\r")
        end_page = input("Max Page = ").rstrip("\r") or 0

    end_page = int(end_page)

    __config__.loadConfig(path)
    artist = __br__.fanboxGetArtistById(id)    
    processFanboxArtist(artist, end_page)
```

bluerthanever:
I get what you meant. But changing it like this doesn't give the guy much time to change the configuration, **IF** the refresh token is changed.

Nandaka:
well, it is a console application. In windows, you can pause the processing by clicking/selecting the console using mouse (if you enable Quick Edit mode), not sure about linux terminal.

bluerthanever:
...still i don't really think the user would be that quick and accurate.
anyway i pulled another request

PR #740, opened 2020-07-16
Dockerfile

DenDen047:
I created `Dockerfile` for PixivUtil2, and wrote the usage in `readme.md`.
Moreover, added ignored files into .gitignore for MacOS.

The docker image is already released on DockerHub.
https://hub.docker.com/repository/docker/denden047/pixivutil2

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,20 @@ numpy-atlas.dll
 tags.txt
 url_list_*.txt
 test.db.sqlite
+.DS_Store
+.AppleDouble
+.LSOverride
+Icon
+._*
+.DocumentRevisions-V100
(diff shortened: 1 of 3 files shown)


Nandaka:
what is the pros and cons running the script directly or using virtual-env compared to docker?

DenDen047:
@Nandaka Thanks for your question. I think docker has some advantages:
- not having any influence to the local env
- easily associating with other programs, such as DB and Machine Learning libraries
- being friendly for users who are not familiar with Python
- having higher replicable

Of course, that has some disadvantages:
- slower processing
- taking the space to save the image data

Anyway, I think more options are favorable for users.

PR #753, opened 2020-08-04
  proper checkUpdatedLimit check 

byjtje:


PixivArtistHandler.py
```diff
@@ -194,7 +194,7 @@ def process_member(caller,
                               PixivConstant.PIXIVUTIL_SKIP_LOCAL_LARGER,
                               PixivConstant.PIXIVUTIL_SKIP_DUPLICATE_NO_WAIT):
                     updated_limit_count = updated_limit_count + 1
-                    if config.checkUpdatedLimit != 0 and updated_limit_count > config.checkUpdatedLimit:
+                    if config.checkUpdatedLimit != 0 and updated_limit_count == config.checkUpdatedLimit:
```
Nandaka, line 197: better use >= incase the updated_limit_count got overflow for whatever reason

PixivTagsHandler.py
```diff
@@ -148,7 +148,7 @@ def process_tags(caller,
                                   PixivConstant.PIXIVUTIL_SKIP_LOCAL_LARGER,
                                   PixivConstant.PIXIVUTIL_SKIP_DUPLICATE_NO_WAIT):
                         updated_limit_count = updated_limit_count + 1
-                        if config.checkUpdatedLimit != 0 and updated_limit_count > config.checkUpdatedLimit:
+                        if config.checkUpdatedLimit != 0 and updated_limit_count == config.checkUpdatedLimit:
```
Nandaka, line 151: same here, better to use >= incase of overflow for whatever reason

byjtje:
Ah okay, was thinking that it might be better to == since there should be no way for it to grow larger than the limit

PR #757, opened 2020-08-07
Update date

byjtje:
Added several date updates, so "daylastupdated" works properly even when not downloading anything.
Though I think it would be best to remove the updating from the updateLastDownloadedImage (unless I am mistaken the last downloaded image value isn't used anywhere?) function


byjtje:
On a side note, no clue why it includes the old commits and a "conflict" which isn't one, sorry for that

Nandaka:
I think you better get the latest commit, merge the changes for this and #758, then only create new pull request?

PR #758, opened 2020-08-07
Regex for the title blacklist

byjtje:
Allowing for better title filtering

### DIFF
--- a/PixivUtil2.py
+++ b/PixivUtil2.py
@@ -1489,6 +1489,10 @@ def main():
             __blacklistTitles = PixivTags.parseTagsList("blacklist_titles.txt")
             PixivHelper.print_and_log('info', 'Using Blacklist Titles: ' + str(len(__blacklistTitles)) + " items.")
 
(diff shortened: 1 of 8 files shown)


PixivUtil2.py
```diff
@@ -1489,6 +1489,10 @@ def main():
             __blacklistTitles = PixivTags.parseTagsList("blacklist_titles.txt")
             PixivHelper.print_and_log('info', 'Using Blacklist Titles: ' + str(len(__blacklistTitles)) + " items.")
 
+        if __config__.useBlacklistTitlesRegex:
+            global __blacklistTitlesRegex
+            __blacklistTitlesRegex = True
+
```
Nandaka, line 1495: I don't see the global `__blacklistTitlesRegex` being used anywhere? It is recommended to access the value from `__config__.useBlacklistTitlesRegex`, as the config might has been changed and reloaded.

I think these lines can be removed.

Nandaka:
also, can you pull the latest commit as it has conflict

PR #779, opened 2020-08-28
Removed unneccessary check from check_version and made the browser op…

byjtje:
…ening opt-in


PixivConfig.py
```diff
@@ -60,6 +60,7 @@ class PixivConfig():
         ConfigItem("Network", "retryWait", 5),
         ConfigItem("Network", "downloadDelay", 2),
         ConfigItem("Network", "checkNewVersion", True),
+        ConfigItem("Network", "openNewVersion", False),
```
Nandaka, line 63: change the default to True, else no one going to be aware of this feature.

If they don't want to use it, they can turn it off by themself

byjtje:
Changed and added it to readme, although I think pop-up like features should always be opt-in

PR #789, opened 2020-09-13
Fixed avatar/background default file extension and background urlfilename

byjtje:
Also added the option for custom background naming formats

### DIFF
--- a/PixivHelper.py
+++ b/PixivHelper.py
@@ -340,7 +340,7 @@ def open_text_file(filename, mode='r', encoding='utf-8'):
     return f
 
 
-def create_filename(artistPage, targetDir, format_src):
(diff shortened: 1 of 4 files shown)


PixivHelper.py
```diff
@@ -340,37 +340,57 @@ def open_text_file(filename, mode='r', encoding='utf-8'):
     return f
 
 
-def create_avatar_filename(artistPage, targetDir, format_src):
-    filename = ''
+def create_filename(artistPage, targetDir, format_src):
```
Nandaka, line 343: possible to change the function name? e.g. make_avatar_bg_filename()

might be confused with make_filename() in the same file:
https://github.com/Nandaka/PixivUtil2/blob/master/PixivHelper.py#L139

Issue #804, opened 2020-10-02
[Feature request] Fanbox functionality for multiple IDs. Request for fanboxlist like option 4


pxssy:
### Prerequisites

* [Y] Did you read FAQ section?
* [Y] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?

### Description

limited fanbox functionality for multiple profile IDs. Not sure if i've exhausted all the possible means, but f2 does not work like 1, where you can dump in  a string of IDs separated by spaces. batch_job doesn't appear to support fanbox options. f4 is possible but manually following every single one on the list is very tedious. 

### Steps to Reproduce

[How option 1 usually works](https://files.catbox.moe/97kl5h.png) Here no artists have ID 1,2,3 so its skipped, nothing wrong here.

[How option f2 is working currently](https://sg.catbox.moe/tgv8mz.png) As you can see, it feeds the input string raw as-is directly into api/url, and not separately as it should in 2. If that's even its intended behavior. 

If possible, can there be a feature that works similarly to option 4, where it takes fanbox ID directly from a fanboxlist.txt?

### Versions

v20200915

bluerthanever:
Doesn't seem too much of a problem. But I have 2 questions:
1. You mentioned option f2, and took option 1 as an example, but you requested a feature like option 4. Just to be clear, do you want a new option like option 4 or want option f2 to be like option 1 or both?
2. I think option 4 is actually more complicated than just "processing from list". If you want a new option like 4, do you want it to be exactly the same, that if you set `processfromdb` to `True`, all processed member ids would be stored in database and be used every time? Also the custom path for each member id, do you also need that?

I think overall making option f2 more like option 1 is easier, and simpler -- no need to add extra options or modify the database.

pxssy:
Option 1 has a parser that first separates the string of IDs into individual job tasks, then sequentially does them. 

It would appear that f2 is missing this parser, f2 only takes in 1 ID and does 1 job. I'm not well versed in coding but it should be able to copy paste the infrastructure from option 1 to first parse the string of IDs? 

In any case, There appears to be a limit of 4096 characters (seems to be bash dependent ) that limits how much IDs can be inputed into option 1. I haven't used option 1 in a long while since option 4 bypasses this character limit.

[...]

PR #805, opened 2020-10-03
Implemented 804

bluerthanever:
Implemented #804 
Main idea:
1. Added new option `listPathFanbox` in `PixivConfig.py`
2. Added support for custom list in `menu_fanbox_download_from_list`, and simply just read list from file, no storing in db or custom paths
3. Added support for multiple creator/user ids in `menu_fanbox_download_by_id`
4. Changed how the above functions uses options (changed from args to option.end_page instead)


bluerthanever:
I used existing option parser options, like end_page, and list_file

bluerthanever:
please don't merge yet...

bluerthanever:
should be good now.

PR #836, opened 2020-10-23
Adding Filename option for Sketch and fixing dates and Unix filenames

byjtje:
__badchars__ were ruining folder / picture names
Fixed default date format, which was creating folders and in a bad format
Changed the json / txt date to be independent of the date format used.

### DIFF
--- a/PixivImage.py
+++ b/PixivImage.py
@@ -407,8 +407,6 @@ class PixivImage (object):
         jsonInfo["Tags"] = self.imageTags
         jsonInfo["Image Mode"] = self.imageMode
         jsonInfo["Pages"] = self.imageCount
-        print(self.worksDateDateTime)
-        print(str(self.worksDateDateTime))
         jsonInfo["Date"] = str(self.worksDateDateTime)
         jsonInfo["Resolution"] = self.worksResolution
         jsonInfo["Tools"] = self.worksTools
--- a/PixivImage.py
+++ b/PixivImage.py
@@ -216,12 +216,12 @@ class PixivImage (object):
 
         # datetime, in utc
         # "createDate" : "2018-06-08T15:00:04+00:00",
-        self.worksDateDateTime = datetime_z.parse_datetime(str(root["createDate"]))
+        self.worksDateDateTime = datetime_z.parse_datetime(root["createDate"])
         # Issue #420
         if self._tzInfo is not None:
             self.worksDateDateTime = self.worksDateDateTime.astimezone(self._tzInfo)
 
-        tempDateFormat = self.dateFormat or "%Y.%m.%d %H:%M"  # 2018.2.27 12:31
+        tempDateFormat = self.dateFormat or "%Y%m%dT%H%M"     # 20180722T1231
         self.worksDate = self.worksDateDateTime.strftime(tempDateFormat)
 
         # resolution
@@ -375,7 +375,7 @@ class PixivImage (object):
         info.write("Tags          = " + ", ".join(self.imageTags) + "\r\n")
         info.write("Image Mode    = " + self.imageMode + "\r\n")
         info.write("Pages         = " + str(self.imageCount) + "\r\n")
-        info.write("Date          = " + str(self.worksDateDateTime) + "\r\n")
+        info.write("Date          = " + self.root["createDate"] + "\r\n")
         info.write("Resolution    = " + self.worksResolution + "\r\n")
         info.write("Tools         = " + self.worksTools + "\r\n")
         info.write("BookmarkCount = " + str(self.bookmark_count) + "\r\n")
@@ -407,7 +407,7 @@ class PixivImage (object):
         jsonInfo["Tags"] = self.imageTags
         jsonInfo["Image Mode"] = self.imageMode
         jsonInfo["Pages"] = self.imageCount
-        jsonInfo["Date"] = str(self.worksDateDateTime)
+        jsonInfo["Date"] = self.root["createDate"]
         jsonInfo["Resolution"] = self.worksResolution
         jsonInfo["Tools"] = self.worksTools
         jsonInfo["BookmarkCount"] = self.bookmark_count
--- a/PixivImage.py
+++ b/PixivImage.py
@@ -221,7 +221,7 @@ class PixivImage (object):
         if self._tzInfo is not None:
             self.worksDateDateTime = self.worksDateDateTime.astimezone(self._tzInfo)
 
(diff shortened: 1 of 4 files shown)


byjtje:
To add to that, using filenameFormat as default for filenameFormatSketch would've been better, but I don't think that's possible without having to rewrite the class

PixivImage.py
```diff
@@ -221,7 +221,7 @@ def ParseInfo(self, page):
         if self._tzInfo is not None:
             self.worksDateDateTime = self.worksDateDateTime.astimezone(self._tzInfo)
 
-        tempDateFormat = self.dateFormat or "%m/%d/%y %H:%M"  # 2/27/2018 12:31
+        tempDateFormat = self.dateFormat or "%Y.%m.%d %H:%M"  # 2018.2.27 12:31
```
jwshields, line 224: I believe if the date format is going to change, it should be changed to conform to ISO8601.
https://en.wikipedia.org/wiki/ISO_8601
There could be multiple ways to format it though..
- Use dashes, or no separators between the dates
- Use a `T` or a space to separate the date from the time, or do not separate them at all
- Use a colon to separate hours from minutes, or do not separate them at all
```
        tempDateFormat = self.dateFormat or "%Y%m%d %H:%M"    # 20182207 12:31
        tempDateFormat = self.dateFormat or "%Y%m%d %H%M"     # 20182207 1231
        tempDateFormat = self.dateFormat or "%Y%m%dT%H:%M"    # 20182207T12:31
        tempDateFormat = self.dateFormat or "%Y%m%dT%H%M"     # 20182207T1231
        tempDateFormat = self.dateFormat or "%Y-%m-%d %H:%M"  # 2018-22-07 12:31
        tempDateFormat = self.dateFormat or "%Y-%m-%d %H%M"   # 2018-22-07 1231
        tempDateFormat = self.dateFormat or "%Y-%m-%dT%H:%M"  # 2018-22-07T12:31
        tempDateFormat = self.dateFormat or "%Y-%m-%dT%H%M"   # 2018-22-07T1231
```
Of this list, I would suggest going with this:
```suggestion
        tempDateFormat = self.dateFormat or "%Y%m%dT%H%M"     # 20182207T1231
```

byjtje: Oh, I wasn't aware that ISO_8601 extended also to the separators of the values, thank you for pointing it out!
As for which format to go with we could also go with just what Pixiv's API returns us directly, giving us something like "2019-08-30T14:51:40+00:00", although I think a shorter solution like yours is better (and colons wouldn't work on Windows)

byjtje:
I implemented the separator change, and got rid of needless functions (the value returned is already a str). Though from Wikipedia, which I somewhat I agree with

> Representations can be done in one of two formats – a basic format with a minimal number of separators or an extended format with separators added to enhance human readability.[14][15] The standard notes that "The basic format should be avoided in plain text."[16] The separator used between date values (year, month, week, and day) is the hyphen, while the colon is used as the separator between time values (hours, minutes, and seconds). For example, the 6th day of the 1st month of the year 2009 may be written as "2009-01-06" in the extended format or simply as "20090106" in the basic format without ambiguity.

But doesn't matter too much tbh

PixivImage.py
```diff
@@ -216,12 +216,12 @@ def ParseInfo(self, page):
...
         # datetime, in utc
         # "createDate" : "2018-06-08T15:00:04+00:00",
-        self.worksDateDateTime = datetime_z.parse_datetime(str(root["createDate"]))
+        self.worksDateDateTime = datetime_z.parse_datetime(root["createDate"])
         # Issue #420
         if self._tzInfo is not None:
             self.worksDateDateTime = self.worksDateDateTime.astimezone(self._tzInfo)
 
-        tempDateFormat = self.dateFormat or "%m/%d/%y %H:%M"  # 2/27/2018 12:31
+        tempDateFormat = self.dateFormat or "%Y%m%dT%H%M"     # 20180722T1231
```
Nandaka, line 224: the date format is configurable in config.ini using `dateformat` and this supposed to be human readable

PixivImage.py
```diff
@@ -375,7 +375,7 @@ def WriteInfo(self, filename):
         info.write("Tags          = " + ", ".join(self.imageTags) + "\r\n")
         info.write("Image Mode    = " + self.imageMode + "\r\n")
         info.write("Pages         = " + str(self.imageCount) + "\r\n")
-        info.write("Date          = " + self.worksDate + "\r\n")
+        info.write("Date          = " + self.root["createDate"] + "\r\n")
```
Nandaka, line 378: same case in here, it supposed human readable.

PixivImage.py
```diff
@@ -407,7 +407,7 @@ def WriteJSON(self, filename):
         jsonInfo["Tags"] = self.imageTags
         jsonInfo["Image Mode"] = self.imageMode
         jsonInfo["Pages"] = self.imageCount
-        jsonInfo["Date"] = self.worksDate
+        jsonInfo["Date"] = self.root["createDate"]
```
Nandaka, line 410: this one is fine, because it should retain the json format

PixivImage.py
Nandaka: also will be converted to local time if it is set to config.ini

byjtje: I'd call "2020-10-12T07:50:09+00:00" easily human readable. For local time we could just use str(self.worksDateDateTime) which would be the same, sans "T" – although not much point in converting since the +00:00 is in there
But which format do you suggest otherwise?

PixivImage.py
byjtje: The standard format is supposed to put in the date, not create random sub folders. Since we are fixing the separators we might as well fix the format, which is barely used anywhere in the world.
I agree with the human readability, but which format would be the best? ISO-like would be 2018-07-22 12:31, with the colons getting replaced in Windows. We could just go for the date without the time instead, which is what I think most people expect from a "work_date"-default anyway.

Nandaka: `2018-07-22` if the use case to create folder based on worksdate is better.

PixivImage.py
Nandaka: use `str(self.worksDateDateTime)` because if `useLocalTimezone` is enabled in the config, then at least it will be converted.

PixivImage.py
jwshields: Ah, I'm sorry, I hadn't realized that this variable was used in other places (I had thought it was only for info/json info) - Sorry about that earlier suggestion.
The colon definitely wouldn't work then.

PR #839, opened 2020-10-24
Fix critical bug introduced by pull #836

byjtje:
Sorry I hadn't fully tested everything yet, since I didn't expect it to get pulled before figuring out the date standards
https://github.com/Nandaka/PixivUtil2/pull/836 tries accessing root when it is part of another method, which is fixed now.
Sketch date is still not working properly which I am going to look at next, but this should be merged asap


Nandaka:
already handled in my local commit, going to push now.

PR #841, opened 2020-10-24
Sketch default date fix

byjtje:
With the default date being this now %works_date_only% is no longer of any use (which it wasn't really before, considering one could just use their own format)


Nandaka:
not sure why shown as pull request, but the code already in the master
image

byjtje:
> not sure why shown as pull request, but the code already in the master
> image

This is Modifying PixivModelSketch.py , not PixivImage.py

Nandaka:
ah, I missed it. I'll update from my side.

PR #851, opened 2020-10-28
Added Series Info and conditionals to JSON/txt, and removed an unneccesary string format

byjtje:
Though about the context of the JSON/txt:
Is there any reason to include Urls, since that one is just a parsing of the Caption? Same could be said for Link.

And for the txt, is there any reason for us to newline with \r\n instead of \n?

### DIFF
--- a/PixivImage.py
+++ b/PixivImage.py
@@ -59,6 +59,7 @@ class PixivImage (object):
     worksDate = ""
     worksResolution = ""
     worksTools = ""
+    seriesNavData = ""
     jd_rtv = 0
     jd_rtc = 0
     # jd_rtt = 0
@@ -196,7 +197,8 @@ class PixivImage (object):
         # title/caption
         self.imageTitle = root["illustTitle"]
         self.imageCaption = root["illustComment"]
-
+        # Series
+        self.seriesNavData = root["seriesNavData"]
         # view count
         self.jd_rtv = root["viewCount"]
         # like count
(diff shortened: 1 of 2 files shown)


Nandaka:
> Is there any reason to include Urls, since that one is just a parsing of the Caption? Same could be said for Link.

this is used for exporting urls to text file if `writeUrlInDescription = True`

> And for the txt, is there any reason for us to newline with \r\n instead of \n?

Windows.

byjtje:
> this is used for exporting urls to text file if `writeUrlInDescription = True`
Not sure if it's intentional, but both Urls and Link get added to txt/JSON with it off. The readme also states a different purpose for writeUrlInDescription

> Windows.
I don't have it on hand to test, but doesn't it still parse them correctly with just \n?

Nandaka:
> I don't have it on hand to test, but doesn't it still parse them correctly with just \n?

No, some windows application (e.g. older notepad), only recognize \r\n combo, https://docs.python.org/3/library/os.html#os.linesep. If you only using one platform, then using os.linesep is OK, but if you open on different platform, it is better to use \r\n so it will shows correctly in windows.

> urls

I think it to make it easier for me when I try copy/paste the link to download manager last time (rather than parsing the whole description manually). Just a convenience only.

byjtje:
Ah I see thanks. For what it's worth the documentation you linked states:
`Do not use os.linesep as a line terminator when writing files opened in text mode (the default); use a single '\n' instead, on all platforms.`

> I think it to make it easier for me when I try copy/paste the link to download manager last time (rather than parsing the whole description manually). Just a convenience only.

Ah, I guess then no reason not to remove them? Or should we keep it at least in the txt?

Nandaka:
> '\n' instead, on all platforms.

Python will write \r\n or \n depending on current platform, right? It might cause an issue if the text file was written in unix-based and then open in Windows.

> json

I think better keep it, someone might uses the json file to import the metadata.

byjtje:
For what it's worth they could still import it, considering the metadata remains in there (and I don't see a use case for the urls specifically)

>It might cause an issue if the text file was written in unix-based and then open in Windows.

Hm for what it's worth googling tells me that it's only really old Windows stuff having problems with it, but I guess it doesn't matter too much.

Just wondering, is there any documentation for the Pixiv API? I've been unable to find anything on their own website

Nandaka:
>  Pixiv API?

I don't think so (or at least in english). I need to do reverse engineering from the current web implementation, other pixiv project, and/or pixiv app.

byjtje:
I see, thanks!

PR #855, opened 2020-10-30
Added series JSON and an option to download it, expended %manga_series% to work with all modes, added an option for untouched JSON

byjtje:
Though the coding is probably not optimal, feel free to leave suggestions.
I also made the %manga_series_x% names work outside of Manga Series mode, and removed %manga_series_desc% from it, which could be included again, but I didn't see a reason to code an exception for something which likely won't get used.
(though I think manga_series -> series would be a better name)


Nandaka:
> (though I think manga_series -> series would be a better name)

There is novel series in pixiv, but I don't have time to implement it yet.

byjtje:
I updated the branch once again to include an option to include an almost untouched JSON from pixiv, to update the readme and remove %work_tools% which did nothing. The removal makes the checking no longer work

Speaking of the JSON, what exactly does the "sl" value represent?
Also not sure if descriptionBooth/YTId are something unique to the artworks, or something general and unrelated like comic/fanboxPromotion.
Also do createDate and uploadDate ever differ?

Nandaka:
> build fails

This is because you remove the `%work_tools%`. Please update the test cases, you can check which unit test failed if you click the details.

> extra stuff in json

Usually I'll just leave those if no one asked. Most of my usage is option 1.

> Also do createDate and uploadDate ever differ?

Yes, sometimes the author will upload a revision without changing the image id (e.g. usually due to insufficient censor bar 😏 )

byjtje:
Ah I see, thanks!
> Usually I'll just leave those if no one asked. Most of my usage is option 1.

Same here, but the ones I removed are stuff which have nothing to do with the image/repetition of stuff already in the JSON:
`"id","title","description","alt","userIllusts","storableTags","zoneConfig","extraData", "comicPromotion","fanboxPromotion"`

I am not sure what exactly descriptionBoothId and descriptionYoutubeId represent, since I have yet to find an artwork with those values set

byjtje:
Done, now the JSON filter is part of config.ini

PixivBrowserFactory.py
```diff
@@ -1019,7 +1020,7 @@ def sketch_get_posts_by_artist_id(self, artist_id, max_page=0):
 
         return artist
 
-    def getMangaSeries(self, manga_series_id: int, current_page: int) -> PixivMangaSeries:
+    def getMangaSeries(self, manga_series_id: int, current_page: int, returnJSON = False) -> PixivMangaSeries:
```
Nandaka, line 1023: can you change the return type from `PixivMangaSeries` to `Union[PixivMangaSeries, str], because this function now can return PixivMangaSeries object or json string (actually getPixivPage also can return BeautifulSoup object, but based on url, it should only return json string)

PixivImage.py
```diff
@@ -150,12 +151,15 @@ def __init__(self,
                 self.originalArtist = self.artist
 
             # parse image
-            self.ParseInfo(payload)
+            self.ParseInfo(payload,writeRawJSON)
```
Nandaka, line 154: missing space here :P I'm using pyflake to lint the code, so it shows up as warning

PixivImage.py
```diff
@@ -528,4 +554,4 @@ def print_info(self):
         PixivHelper.safePrint(f'Pages          : {self.current_page} of {int(self.total_works/works_per_page)}')
         PixivHelper.safePrint('Works          :')
         for (work_id, order) in self.pages_with_order:
-            PixivHelper.safePrint(f' - [{order}] {work_id}')
+            PixivHelper.safePrint(f' - [{order}] {work_id}')
```
Nandaka, line 557: same here, pyflake also ask for newline at the end of file

byjtje:
All done!

Nandaka:
Thanks a lot 👍

byjtje:
Thank you too :)

byjtje:
Just in case it is of any use for anyone
> I am not sure what exactly descriptionBoothId and descriptionYoutubeId represent, since I have yet to find an artwork with those values set

Those are values which get parsed from the artwork's description when linking a Booth item / Youtube video. Though it seems to only take the first value it gets, i.e. https://www.pixiv.net/artworks/85208671 returns only "bRCvFd4VGlY"

PR #857, opened 2020-11-03
fix bad standard config value

byjtje:


Nandaka:
better call .strip() to remove the whitespace @ https://github.com/Nandaka/PixivUtil2/blob/master/PixivImage.py#L410-L411.

I think user might configure with space after the comma.

byjtje:
Done

PR #867, opened 2020-11-16
Allow pasting of urls for input

byjtje:


PixivHelper.py
```diff
@@ -553,18 +553,21 @@ def have_strings(page, strings):
...
+            if len(temp) > 0:
+                try:
+                    _id = temp
+                    ids.append(_id)
+                except ValueError:
+                    print_and_log('error', u"ID: {0} is not valid".format(id_str))
+    else:
+        ids = re.findall("(?:series|users|\s|,|^|artworks)\/?(\d+)", ids_str)
+        if not ids:
+            print_and_log('error', u"ID: {0} is not valid".format(ids_str))
```
Nandaka, line 570: need to change the logic because is_string is not always set to True.
image

I think you better add a check if the ids_str is start with http or not, e.g. :
``` python
if ids_str.startswith("http"):
    # check by regex
else:
    # original logic

```

Also need to take care if it is mixed case (url with normal id)?

byjtje: > need to change the logic because is_string is not always set to True.
> image

Ah yes, it's meant for the case when `is_string=False` (so the case when we don't paste the fanbox/sketch member id/name, which would be something for the if `is_string=True` case). Thinking of it, the `is_string=True` case is also very easy to handle with urls, so I am going to add that I guess.

> I think you better add a check if the ids_str is start with http or not, e.g. :
> Also need to take care if it is mixed case (url with normal id)?

It does already that with the regex 
`ids = re.findall("(?:series|users|\s|,|^|artworks|posts)\/?(\d+)", ids_str)`
checking for numbers which are prepended by certain stuff, so we don't accidentally catch stuff like the page number or artist id when downloading series

byjtje:
All done now, should also work for fanbox/sketch ids now, and removed the now redundant seperator

byjtje:
And fixed the latest test fail, apparently the "Change all Occurrences" in vscodium doesn't work as you'd expect it to ^^

bluerthanever:
Hmm. Just tested the new `get_ids_from_csv` like this:
> import PixivHelper as ph
> string = "https://www.fanbox.cc/@g-ondom"
> ph.get_ids_from_csv(string, True)
the result is `['g']` instead of `['g-ondom']`.
same result when string = 'g-ondom'

PixivHelper.py
```diff
@@ -552,19 +552,16 @@ def have_strings(page, strings):
...
-                _id = temp
-                if not is_string:
-                    _id = int(temp)
-                ids.append(_id)
-            except ValueError:
-                print_and_log('error', u"ID: {0} is not valid".format(id_str))
+def get_ids_from_csv(ids_str, is_string=False):
+    ids = []
+    if is_string:
+        ids = re.findall("(?:@|^|https:\/\/(?!www\.)|\s|,)(?!https:)(\d+|\w+)", ids_str)
```
Nandaka, line 558: need to add additional keyword after \W+ then?

bluerthanever: I guess so. I changed the regular expression to `(?:@|^|https:\/\/(?!www\.)|\s|,)(?!https:)(\d+|\w[\w-]*\w)`. Seems to be working for '-'s. But  not sure if this is the best one....

Nandaka:
@byjtje 

found another issue for pixiv sketch:
https://sketch.pixiv.net/@user_rehz3537, looks like it match the 'sketch' string too.

Maybe need to modify to `(?:@|^|https:\/\/(?!www|sketch\.)|\s|,)(?!https:)(\d+|\w[\w-]*\w)`

PR #874, opened 2020-12-03
Fixes for lots of stuff, new features and download improvements

byjtje:
~~This is still very incomplete, and some stuff is removed for no real reason, so please don't mind that.~~
Everything should be implemented and working now (hopefully)

I was thinking of changing the Artist handler to work a bit differently when using blacklists / datediff, as we can use the API that gets called for pages when browsing normally, to get everything we need to filter up to 100 images at once and thus save a lot of time for people like me whose filters apply a lot. However the downside is that it makes an extra call for when it doesn't filter – which normally shouldn't be a problem, but is almost guaranteed when downloading an artist already gone through (Download from list).
The only way I can think of with which we can somewhat avoid this is by checking the image ids beforehand (especially with checkUpdatedLimit) and then seeing if there even is anything to download. However that would only work when the redownload settings are disabled. So I guess it being opt-in would be the best way to go about it with it checking that those settings are off, but does anyone have a better idea?

 

Additionally this implements a way to download between IDs, lets one download bookmarks with URL, and fixes faulty bookmark downloading (wrong artist for naming / downloads cover+bg / R18 mode being ignored)


byjtje:
Added an option to also download artist JSON, however no matter what I try, it seems to not converted to unicode

Nandaka:
Usually I don't use filter at all, and just download everything.

Rather than overwriting the default handler, I think it is better if you create your own handler in separate file, then just add the new menu from the main script for your use case. This way it can support both your and my use case.

Also maybe change the menu title from `14. Download by member_id and image_id` to `14. Download by member_id (preprocess image id)` (or something else).

Nominally, how much is the difference and does it return the same result?

byjtje:
I found it easier to just rewrite it, since it also made implementing the ID download (which is almost exactly the same as normal artist downloads after all) and fixing the issues with bookmarks easier. Apparently any tag filtering for bookmarks doesn't work properly from the URL. (although one can ask how important the bookmark DL option is)

It already supports both use cases, and with the only difference being when blacklisting is enabled. I can just add a config option to skip the new handling, although I was hoping that there would be another way, thus asking.

`Nominally, how much is the difference and does it return the same result?`
It takes ~1.5-2s per AJAX request here. So each one saved saves around that much time. It depends on the artist, but when downloading some fully it can easily reach several minutes

Also unrelated to this, , I fixed paths for `parseTagsList` functions I added a function to only log the `image_id` being skipped, to make re-downloading more easily, however it needs checking on whether or not anything gets missed.

byjtje:
>Also maybe change the menu title from 14. Download by member_id and image_id to 14. Download by member_id (preprocess image id) (or something else).

14. is for downloading a members images not by pages, but by ID, and a separate change from the preprocessing

Nandaka:
I prefer to have separate files, so it is easier to fall back, rather rolling back all the changes inbetween. If it is stable enough, then I can just replace the whole thing

byjtje:
> If it is stable enough, then I can just replace the whole thing

Ah yes, that is ultimately the goal, this build isn't ready to be merged yet. I was just hoping to get some feedback on the logic first

byjtje:
@Nandaka It should (hopefully) be now in a usable state. Let me know if you find something, if not I can make it mergeable.

Only things needing fixing are (once again hopefully):
Artist.json not being encoded in Unicode (I'd need help on that, I cannot find the reason / a fix), download by tag stopping if a whole page has been skipped by blacklists (will fix that tomorrow), and maybe some `print`s being left over / still commented out, R-18 mode not working on R-18G (I'd first need to know all the variations of it – I don't use this for R-18 so I got no clue on them, sorry). opvalid stuff is still untested.
Also with `preprocess=2` images get checked twice from the DB (I will check that tomorrow,  alongside db.cleanupFileExists (is there a need to check fileextensions, especially for every image type?)

If one wants they could also implement a new way to download an artist by several tags with the new `process_blacklist` function.

byjtje:
Changed BatchHandler, now it makes a copy of the config and lets us change every value. Also fixes https://github.com/Nandaka/PixivUtil2/issues/880

byjtje:
Fixed backslashes being replaced on Unix, and made slash be replaced to ／. Could also ⧸ or maybe other characters, any thoughts on it?

Nandaka:
> Just wondering, when exactly does one get duplicated results in tag search? I couldn't find any JSON calls which did in testing, and removing the parts about it makes no difference (keeping them in wouldn't do any either I guess).

last time it might happen when someone upload a lot of images with the same tags, so the current page is shifted by 1 page.

> Fixed backslashes being replaced on Unix, and made slash be replaced to ／ on it. Could also use ⧸ or maybe other characters, or even implement it in Windows, any thoughts on it?

It won't work if the file system doesn't support unicode (e.g. FAT32).

> Also integrated sanitize_filename into make_filename, since every call of it, outside of tests, included it afterwards. Tests are the only reason as to why targetDir is optional (too lazy to rewrite them :> )

You need to update the unittest, because I use it to check if the parser working correctly or not. Normally I will split the function rather than creating big function that hard to read, also to make it easier to test. You need to think maintenance.

PixivBatchHandler.py
```diff
@@ -201,7 +183,7 @@ def process_batch_job(caller: PixivUtil2, batch_file=None):
 
     if os.path.exists(batch_file):
         jobs_file = open(_default_batch_filename, encoding="utf-8")
-        jobs = demjson.decode(jobs_file.read())
+        jobs = demjson.decode(jobs_file.read()) #is there a reason to use demjson over json?
```
Nandaka, line 186: I remember that pixiv has a json that out-of-spec, where the key is an integer, not a string.

For this instance, I think we can use built-in json lib

byjtje:
It should hopefully be ready to merge now, latest changes from the main branch were included.

> last time it might happen when someone upload a lot of images with the same tags, so the current page is shifted by 1 page.

Oh, I was wondering for https://github.com/Nandaka/PixivUtil2/blob/e374a7aa9b2b8548cbbb8bb69e565ae23ddabe78/PixivTagsHandler.py#L83 . I since I didn't see a reason to keep it in

> You need to update the unittest, because I use it to check if the parser working correctly or not. Normally I will split the function rather than creating big function that hard to read, also to make it easier to test. You need to think maintenance.

Ah I didn't merge them, just made it automatically sanitize if we include the directory. But best to keep both options, I forgot about the potential uses with avatar/bg names.
I updated testSanitizeFilename3 to use the new make+sanitize. 
Though it should be worth mentioning that fixing backslashes also makes \\ no longer count as / for Unix.

byjtje:
@Nandaka 
Tested opisvalid stuff now, should be ready to merge. Only issues I found are ones present in the main branch (most opvalid options seem to restart when pressing 'x' in the menu, instead of letting one exit the program, and the avatar url changing not working for some artists, causing 404s)

byjtje:
@Nandaka Sorry to ask, but is there anything that still needs fixing in order for it to get merged?

Nandaka:
> @Nandaka Sorry to ask, but is there anything that still needs fixing in order for it to get merged?

I'm still trying to understand the changes, sorry about that. (a bit busy with real job...)

PixivArtistHandler.py
```diff
@@ -12,33 +12,28 @@
...
 import PixivImageHandler
 from PixivException import PixivException
+from PixivListHandler import process_blacklist, process_list_with_db
 
 
 def process_member(caller,
                    config,
                    member_id,
-                   user_dir='',
                    page=1,
```
Nandaka, line 21: Why `user_dir` is removed? It is used for option 4 (download from list), as it can define the custom path for each member.
image

PixivImageHandler.py
```diff
@@ -23,15 +23,14 @@ def process_image(caller,
                   config,
                   artist=None,
                   image_id=None,
-                  user_dir='',
```
Nandaka, line 26: same here, as we need to pass the customized save folder from list.tx

byjtje:
> > @Nandaka Sorry to ask, but is there anything that still needs fixing in order for it to get merged?
> 
> I'm still trying to understand the changes, sorry about that. (a bit busy with real job...)

I didn't mean to be imposing, sorry if it came across that way!

Sorry about that user_dir, I must have somehow missed that functionality being a thing. Is it intended to only work for the list download, or also for downloading single images by id?

I will update PixivListItem.parseList alongside, as it currently does not work properly for unix absolute path and new member URLs

Nandaka:
> Sorry about that user_dir, I must have somehow missed that functionality being a thing. Is it intended to only work for the list download, or also for downloading single images by id? (it currently doesn't do so for the latter)

it is only from list.txt.

on side notes, PR with a lot of feature changes (not the line of codes) is harder review, so normally I prefer separate PR for each feature/bug fix changes.

byjtje:
user_dir should be fixed now. Although seeking feedback on what to do with non-Windows paths. It's fixed, but could change stuff if anybody on Linux used Windows syntax. However making it an optional setting seems wrong to me. Maybe include a warning for one or two versions if it detects \ in there?

> on side notes, PR with a lot of feature changes (not the line of codes) is harder review, so normally I prefer separate PR for each feature/bug fix changes.

Going to do so in the future, sorry

Nandaka:
hmm, I think it doesn't do the testing because it has conflict? I notice it does works for other pull request.
image

image

byjtje:
Fixed a bad way of getting ceiling which wouldn't work with multiples of 100, and included the throw_empty_error=True from main branch.
@Nandaka If there are still certain things needing fixing please tell me, although this is so far behind main now ^^

Nandaka:
I think it is too many changes, that I cannot understand it already.

cglmrfreeman:
This PR is now _very_ behind master and has absolutely no description added per commit. @byjtje, were you still pursuing this or is it now abandoned?

holopekochan:
The PR does not have a good commit message.
Better to rebase the commit, it is hard to check.

Issue #896, opened 2021-01-19
Suspended Fanbox artist causing Fanbox download functionalities to stop; artist also isn't actually suspended


Zekira:
### Prerequisites

* [Y] Did you read FAQ section in [readme.md](https://github.com/Nandaka/PixivUtil2/blob/master/readme.md)?
* [Y] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?
* [Y] Did you search for existing issues in [Issues](https://github.com/Nandaka/PixivUtil2/issues)?

### Description

Related, maybe even same as https://github.com/Nandaka/PixivUtil2/issues/873

Artist ID is 43189578. Their fanbox is working fine. (nsfw warning).

Should also be noted that their non-fanbox has a different ID for some reason(58633620).

### Steps to Reproduce

1. Use any of the f# functions. f2 should be fine.
2. In Artist ID, input 43189578.
3. Error shown:
Error getting FANBOX artist by id: 43189578 ==> 該当ユーザーのアカウントは停止されています。
Unknown Error: (<class 'UnboundLocalError'>, UnboundLocalError("local variable 'artist' referenced before assignment"), <traceback object at 0x03F03408>)

**Expected behavior:** Should get the artist successfully, or if the user really is suspended, they shouldn't be included in the retrieval of the following list, or skip over it instead and don't close the program so that batch functions like f1, f4, and f5 don't break in between.

**Actual behavior:** Error causes program to exit despite user existing

### Versions

20210108-beta4

You can get this information from executing `PixivUtil2.py --help`.
Latest version available in https://github.com/Nandaka/PixivUtil2/releases

bluerthanever:
it was somehow changed in #873 to: if you have downloaded from their pixiv and their information is restored in DB before, PixivUtil2 would use the stored information in db to generate filenames. otherwise the error would be thrown as before

the reason for keeping throwing such error is because if some filename format might contain placeholders whose information could only be retrieved from pixiv.net, namely artist name and token, filenames could be wrong.

i would change it to continuing processing next artist instead of stopping the whole program though. that was not though out.

Zekira:
> it was somehow changed in #873 to: if you have downloaded from their pixiv and their information is restored in DB before, PixivUtil2 would use the stored information in db to generate filenames. otherwise the error would be thrown as before
> 
[...]

Issue #889, opened 2020-12-30
Support File Renaming to Remove "Invalid" Characters


srvrguy:
### Prerequisites

* [x] Did you read FAQ section in [readme.md](https://github.com/Nandaka/PixivUtil2/blob/master/readme.md)?
* [x] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?
* [x] Did you search for existing issues in [Issues](https://github.com/Nandaka/PixivUtil2/issues)?

### Description

Images are currently downloaded with the title of the artwork as part of the file name. For some systems, this means it tries to save files with invalid characters for some file systems or services. For example, OneDrive doesn't support the filename "86475500_p0 - 紫罗兰永恒花园 外传礼服薇～(*_ω\\\*).jpg" but will work with the asterisks removed.

It would be useful to have a configuration setting that contains characters considered "invalid" and that the program can strip from the directory or file name when saving. It doesn't even need to be enabled by default, but just there for this type of use  case.

### Steps to Reproduce

1. Download an image that has a character considered invalid

**Expected behavior:** The file is saved without invalid characters in the name

**Actual behavior:** The file is saved with invalid characters

### Versions

20201212-beta3

Likely others as well.

bluerthanever:
1. Whether to include title of the artwork as part of the filename can be controlled with a whole lot of filename format options under filename in configuration. If you want to exclude the title, simply remove `%title%` or change the format to whatever you want.
2. Asterisks **ARE** removed when files are saved locally. Take the artwork in your description for example, the filename was changed into `86475500 - 紫罗兰永恒花园 外传礼服薇～(__ω＼_)` locally, with asterisks replaced with underline.
No idea what you are using but the PixivUtil2 I have been using always does the above.

byjtje:
>     2. Asterisks **ARE** removed when files are saved locally. Take the artwork in your description for example, the filename was changed into `86475500 - 紫罗兰永恒花园 外传礼服薇～(__ω＼_)` locally, whose asterisks were replaced with underlines.
> 
> 
> No idea what you are using but the PixivUtil2 I have been using always does the above.

Only on Windows

bluerthanever:
Oh. Just noticed that the bad characters are indeed different. Guess no one took care of that before.

byjtje:
[...]

PR #897, opened 2021-01-19
Fix

bluerthanever:
Mainly:
1. Fixed when downloading fails to be saved to desired path but under the folder of PixivUtil2, desired path instead of actual path is saved to db, as spoken of in https://github.com/Nandaka/PixivUtil2/issues/889#issuecomment-761727105 under issue #889 
2. Optimized logic for FANBOX artists who are suspended on pixiv, as in #896 , also fixed a bug I caused and didn't realize.

Did not implement the `custom bad characters` idea, because of laziness, and I have to agree with @byjtje , that it's not PixivUtil2's problem, and it could be solved with other tools, like batch renaming tools.
I am not so sure about the actual file path thing... but whatever, I never use that column or related menus.

### DIFF
--- a/PixivDownloadHandler.py
+++ b/PixivDownloadHandler.py
@@ -206,10 +206,10 @@ def download_image(caller,
                 # write to downloaded lists
                 if caller.start_iv or config.createDownloadLists:
                     dfile = codecs.open(caller.dfilename, 'a+', encoding='utf-8')
-                    dfile.write(filename + "\n")
+                    dfile.write(filename_save + "\n")
                     dfile.close()
 
-                return (PixivConstant.PIXIVUTIL_OK, filename)
+                return (PixivConstant.PIXIVUTIL_OK, filename_save)
 
             except urllib.error.HTTPError as httpError:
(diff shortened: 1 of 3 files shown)


bluerthanever:
well, if you agree to implement the `custom bad characters` thing, or the check download history for `download from bookmarked artists`, i could also try to do that. don't have anything to do anyway.

Nandaka:
> the filename stuff

It think it was artifact of Issue #548

> custom bad chars stuff

I guess it should be fine? Just need to change https://github.com/Nandaka/PixivUtil2/blob/master/PixivHelper.py#L94 to read from config if config is not empty?

bluerthanever:
> It think it was artifact of Issue #548

not so sure about that, if this is about cglmrfreeman's situation... the test in that issue i asked him to try passed... and in 548 it was clearly unicode error, but in 889 it was oserror, and i tried to look up error code 22 in linux, and could not find really detailed explanation.

>I guess it should be fine?

hmm, are you going to do it? =)
i was thinking of a different approach though, but now i think yours is better. haha.

Issue #899, opened 2021-01-26
Improve FANBOX HTML output


LizeHelesta:
### Prerequisites

* [x] Did you read FAQ section in [readme.md](https://github.com/Nandaka/PixivUtil2/blob/master/readme.md)?
* [x] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?
* [x] Did you search for existing issues in [Issues](https://github.com/Nandaka/PixivUtil2/issues)?

### Description

0. The sample post: [726146](https://www.fanbox.cc/@mataro/posts/726146)

1. FANBOX HTML output doesn't process the styles of block p.

    Post Info:

    ![Snipaste_2021-01-26_18-10-00](https://user-images.githubusercontent.com/61379118/105831486-bc944700-6001-11eb-841a-a34583320ef0.png)

    Original Post Appearance:

    ![Snipaste_2021-01-26_18-08-57](https://user-images.githubusercontent.com/61379118/105831420-a8e8e080-6001-11eb-93b6-36d56d937085.png)

    HTML output:

    ![Snipaste_2021-01-26_18-29-28](https://user-images.githubusercontent.com/61379118/105833719-8dcba000-6004-11eb-973c-0fdb566c4c3d.png)

    I had try to solve it by myself:
    ```python
                        else:
                            block_text = block["text"]
                        ### Start
                        if "styles" in block:
                            for i in reversed(range(0, len(block["styles"]))):
                                style = block["styles"][i]
                                if style["type"] == "bold":
                                    style_offset = style["offset"]
                                    style_length = style["length"]
                                    block_text = block_text[:style_offset+style_length] + "</b>" + block_text[style_offset+style_length:]
                                    block_text = block_text[:style_offset] + "<b>" + block_text[style_offset:]
                                else:
                                    raise PixivException(f"Unknown style", errorCode=PixivException.OTHER_ERROR)
                        ### End
                        self.body_text = f"{self.body_text}<p>{block_text}</p>"
                    elif block["type"] == "image":
    ```

2. `BeautifulSoup.prettify()` will remove ideographic space `\u3000` which is important to keep the post layout neat in some japanese post.

    Original Post Appearance:

    ![Snipaste_2021-01-26_18-14-11](https://user-images.githubusercontent.com/61379118/105832513-0598cb00-6003-11eb-8751-b67a26291eaf.png)

[...]

PR #902, opened 2021-01-28
Fix #889 #899

### LINKED ISSUE #889: Support File Renaming to Remove "Invalid" Characters

bluerthanever:
For details please refer to each commit.
Mainly:
1. implemented `customBadChars` and added description in `readme.md`, may not work well with spaces
2. optimized methods to parse fanbox posts and the html template and the styles in it.


bluerthanever:
for the link and styles, the logic i wrote may be a little complicated? but it should be able to deal with such situations, even though no one would probably set the texts like this, but just in case:
original:
original
parsed:

bluerthanever:
for the link and styles, the logic i wrote may be a little complicated? but it should be able to deal with such situations, even though no one would probably set the texts like this, but just in case:
original:
original
parsed:
parsed

bluerthanever:
hmm.. i think i am gonna change the css styles a little more...

PixivModelFanbox.py
```diff
@@ -255,8 +310,7 @@ def parseBody(self, jsPost):
                 elif block["type"] == "embed":  # Implement #470
                     embedId = block["embedId"]
                     if embedId in jsPost["body"]["embedMap"]:
-                        embedStr = self.getEmbedData(jsPost["body"]["embedMap"][embedId], jsPost)
-                        self.body_text = f"{self.body_text}<br />{embedStr}"
+                        self.body_text += self.getEmbedData(jsPost["body"]["embedMap"][embedId], jsPost)
```
LizeHelesta, line 313: Sequence `embed` block will wrap to one line, when the author link plural posts at once.

bluerthanever: hmm. how about we just wrap it up with a `p` tag. everything would be wrapped in `p`. i think that would be neat.

bluerthanever: i will do some tests....

bluerthanever: hmm, just tested with posting a post.
1. for images and files, during editing, even if i deleted the empty line between 2 continuous files or images or one of each, after publish, there is still a blank `p` between them.
2. not the same with embed, no empty `p`s between or before them.

i think i will just wrap them in `p` tags, if you don't mind?

bluerthanever: sorry, earlier reply was wrong. please ignore.
just tested with posting a post.
1. for images, during editing, even if i deleted the empty line between 2 continuous images, after publish, there seem always to be a blank p between and after them.
2. not the same with embeds or files, no empty ps between or before them.

i think i will just wrap them in p tags, if you don't mind?

bluerthanever:
Now it should be like this:
original:
QQ图片20210129161315

parsed:
Screenshot_2021-01-29 よりアオイ - This is another test

bluerthanever:
why all of a sudden all the test scripts are picking one me....? 
i am gonna hit the gym and fix this later. please comment or anything.

bluerthanever:
hmm. this is weird. i will pull a new one...

PR #903, opened 2021-01-29
Fix #889 #899

### LINKED ISSUE #889: Support File Renaming to Remove "Invalid" Characters

bluerthanever:
For details please refer to each commit.
Mainly:
1. implemented `customBadChars` and added description in readme.md, may not work well with spaces
2. optimized methods to parse fanbox posts and the html template and the styles in it

for the link and styles, the logic i wrote may be a little complicated? but it should be able to deal with following situations, even though no one would probably set the texts like this, but just in case:
original:
original
parsed:
parsed

for article posts with images, files and embeddings, the result should be like this:
original:
QQ图片20210129161315

parsed:
Screenshot_2021-01-29 よりアオイ - This is another test

hopefully this request would pass.


bluerthanever:
hmm. this is confusing now. here on my side it says the template didn't pass. but how? it's just a template, and i don't see any test scripts using it. do i also need to update all the test cases? like run `test_updateHtml.py` and the update all the files under `.\test`?

LizeHelesta:
The pull request test build will be triggered when you add commits at once, so the check failed didn't blame to the last commit "Update template.html" (9ade0411e7cd422252d98722d9c4e5081973622a) but all added commits at once.
The last commit (4586ef3616c83b1c27163dacb47450cfec4288d9) you added just fix one check and still two check cannot pass because the length of HTML output body text has been changed after refactoring the block handler.
You can check the detail and fix them. https://travis-ci.com/github/Nandaka/PixivUtil2/builds/215245670

bluerthanever:
oh.. i could not open that page, and now i am accessing with proxy. i thought there were no details. thanks.

bluerthanever:
weirdly, i swear that when i ran ` test_PixivModel_fanbox.py` there was just 1 error.

bluerthanever:
not sure why yet, but the `@media` part of the css doesn't seem to work on chrome and ie. checking again tomorrow... late now.

bluerthanever:
@LizeHelesta hmm. sorry that i did another run of test, and it seems like there's something wrong with the test earlier, 
as for `p`, `image`, `file`, and `embed`, between any combination of these four, or two of each kind, there aren't necessarily an empty `p`. so... i guess i have to rethink again...

bluerthanever:
never mind for the last deleted reply... i was overthinking....

Nandaka:
fyi, travis ci running on linux (because I don't have one). Usually I run the test in my windows pc before pushing the commit to origin, so I need travis ci to check for linux box.

bluerthanever:
usually i ran with creators i am supporting.

bluerthanever:
ok, should be done now. fixed the css problem not working on chrome and ie.

bluerthanever:
hmm. can you wait for a bit. i wanna test something more... still not sure about the blocks.

bluerthanever:
okay, finally i think this is it...

if util runs into continuous images that are too small, my commit (removing `<br />` before each image block) would cause all images go together. now i just use css to control that... limit the maximum width of the most outside container/div, then make each image's width to 100% to make sure each image fills up one row....

bluerthanever:
okay... i think this is finally it now.

before i was kinda not sure about removing the `<br />` before each image block, now i am, and i am using css to make sure each image fills up one row/line.

LizeHelesta:
`img` should not force `width: 100%`, it isn't reasonable for small size images.
`.cover` and `.title` may be need `margin-bottom` to make them more like original layout.

bluerthanever:
hmm. i agree that some small images get stretched and look kinda weird. i think i will remove that style  and think of another way. and adjust a little more.
but you know this is just a template and you could write your own with your own css right?

bluerthanever:
gee... the more i adjust, the more i wanna close this request and start over again...

bluerthanever:
this should be good enough now i guess? css is so new to me...

LizeHelesta:
I think it is good enough. And the gap between title and cover or body text which makes layout intuitively clear is a basic point.
I agree to re-open a new pull request to keep commits streamlined and easy-to-read. Because you have talk about this, i also want to understand why you not edit files in batches rather than editing every single file via web?

bluerthanever:
usually i would submit after all i have done completely as well.. but you know, there could be uncertain variables, or that you just lie down and go over the changes in your mind, and think that some changes could be done with better ways... so you got up and redo them.
btw i am not using git. i don't really have a repository so not really using it to keep track.
@Nandaka do you prefer that i should reopen a clean RP? i don't mind though.

bluerthanever:
i think i just found another not so "big-deal" thing, that, now, since i used `display:grid` for `.caption`, when the image is small or not wide enough, the `a` tag actually fills up the whole row/line, which means if you click the white/empty area on both side of the image, you get the same result as when you click on the image,
do you want that fixed?

bluerthanever:
i decided to just open a new one... to fix the image hyperlink issue. i will pull a clean request later, or maybe tomorrow

bluerthanever:
sorry... just found out the fix was easy, just add `a{margin:auto;}` and it's done... so i reopened this... tell me if i need to pull a new one. whew.... off to bed now.

PR #919, opened 2021-02-26
Bad chars enhancement

bluerthanever:
1. added new class to realize customized bad char replacement
 - 2 hidden fields for output string and regular expression replacement
 - initialization method to parse `customBadChars` string
 - method to actually do the replacement
2. added `followup` parameter for option `customBadChars`
3. changed codes to call the newly added method
4. added description for the new syntaxes for bad chars.

### DIFF
--- a/PixivConfig.py
+++ b/PixivConfig.py
@@ -158,15 +158,12 @@ class PixivConfig():
         ConfigItem("Settings", "setLastModified", True),
         ConfigItem("Settings", "useLocalTimezone", False),
 
-        ConfigItem("Filename",
-                   "filenameFormat",
+        ConfigItem("Filename", "filenameFormat",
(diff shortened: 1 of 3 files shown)


bluerthanever:
sorry that i added too many commits. sometimes i just wanted to make it better but made some mistakes which i found out later... and had to fix them which might leads to more mistakes or dissatisfaction, so i had to change more...

PixivConfig.py
```diff
@@ -14,6 +15,65 @@
 script_path = PixivHelper.module_path()
 
 
+class CustomSanitizer:
```
Nandaka, line 18: better move it to separate file or merge it with pixivHelper as this is not configuration.

bluerthanever: hmm... okay, i will pull a new one in that case, this is kinda messy now anyways.

PR #944, opened 2021-05-06
fixed an ANSI escape code

DenDen047:
the error of ANSI escape code occurred when the program is run using `docker-compose up`

### DIFF
--- a/PixivHelper.py
+++ b/PixivHelper.py
@@ -345,7 +345,7 @@ def set_console_title(title):
         except FileNotFoundError:
             print_and_log("error", f"Cannot set console title to {title}")
     else:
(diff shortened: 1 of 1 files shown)


cglmrfreeman:
@DenDen047 This fix is causing escape code issues with my distro as well as an Ubuntu virtual machine I tested it. See issues #947 and #948. Did you test this fix outside of docker before commit?

Also, for the record `set_console_title` has never worked in my distro and apparently does not work in Ubuntu either as per screenshots.

whinette:
I've proposed a patch here; https://github.com/Nandaka/PixivUtil2/pull/950
This will also fix the bell issue that was caused by `\x07`: https://charbase.com/0007-unicode-bell

Issue #941, opened 2021-04-28
Download resume after disk space error skips download in progress rather than retrying


nomakewan:
### Prerequisites

* [✔] Did you read FAQ section in [readme.md](https://github.com/Nandaka/PixivUtil2/blob/master/readme.md)?
* [✔] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?
* [✔] Did you search for existing issues in [Issues](https://github.com/Nandaka/PixivUtil2/issues)?

### Description

If in the process of downloading from Fanbox the destination disk runs out of space, PixivUtil2 will correctly stop attempting to write to the disk and will display a "Press enter to retry" button waiting for the user to correct the low space condition. However, upon pressing enter, PixivUtil2 will skip over the file it was attempting to write rather than reattempt the download, thus causing a missed image. I only (accidentally) tested this while downloading from Fanbox and do not know if it applies to regular Pixiv as well.

### Steps to Reproduce

1. Have remaining hard drive space less than required to download entire Fanbox contents.
2. Attempt to download Fanbox contents.
3. When remaining space hits the limit, PixivUtil2 will pause awaiting user input in the middle of a download.
4. Correct the space issue, then press enter to retry.

**Expected behavior:**
PixivUtil2 will redownload the image it was in the process of downloading before the destination disk ran out of space.

**Actual behavior:**
PixivUtil2 skips the image it was in the process of downloading and downloads the next image in the list.

### Versions
20210421

bluerthanever:
https://github.com/Nandaka/PixivUtil2/blob/24503c404269bb47528b26093bd72c69d6f5fcf9/PixivDownloadHandler.py#L228
change here to `continue`?

Nandaka:
should be working, as it still within the while-loop:
https://github.com/Nandaka/PixivUtil2/blob/24503c404269bb47528b26093bd72c69d6f5fcf9/PixivDownloadHandler.py#L59

nomakewan:
I have attached the log output from the particular section (and scrubbed the filenames). The issue was that I left PixivUtil2 running and went to sleep. When I woke up it was waiting for my input, and it appears that it broke out of that while-loop waiting for my input. The issue is that if it's waiting for user input, it shouldn't be able to break from that loop (or should at least go back to the start of the loop after user input).

[...]

PR #945, opened 2021-05-07
Update PixivDownloadHandler.py

bluerthanever:
Fixing #941


bluerthanever:
just a simple change.

also i am thinking about changing `print_and_log`, and where it's called. but it seems too much work....

Issue #948, opened 2021-05-08
While PixivUtil2 is running in the console in the background, it triggers KDE "Bell in session 'Default" notifications


cglmrfreeman:
### Prerequisites

* [X] Did you read FAQ section in [readme.md](https://github.com/Nandaka/PixivUtil2/blob/master/readme.md)?
* [X] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?
* [X] Did you search for existing issues in [Issues](https://github.com/Nandaka/PixivUtil2/issues)?

### Description

PixivUtil2 while running in the background in a terminal window causes "Bell in session 'Default'" notifications, or an [XBell signal](https://www.x.org/releases/X11R7.5/doc/man/man3/XBell.3.html). This means that PixivUtil2's script is somehow creating a [Bell Character](https://en.wikipedia.org/wiki/Bell_character).

![Screenshot_20210507_200644](https://user-images.githubusercontent.com/9744201/117524074-c28ab400-af70-11eb-8e46-d9da0b396414.png)

### Steps to Reproduce
(Running a Linux distribution with KDE and using Konsole)
1. Download by Member ID. It seems most Member IDs trigger the notification multiple times while processing Image IDs (both downloading or detecting if they are already downloaded)

**Expected behavior:**
PixivUtil2 not to trigger any notifications.

**Actual behavior:**
PixivUtil2 is triggering the XBell signal somehow.

### Versions

PixivDownloader2 version 20210421

Issue #947, opened 2021-05-08
Strange Typo "ixivDownloader" before constant PIXIVUTIL_VERSION


cglmrfreeman:
### Prerequisites

* [X] Did you read FAQ section in [readme.md](https://github.com/Nandaka/PixivUtil2/blob/master/readme.md)?
* [X] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?
* [X] Did you search for existing issues in [Issues](https://github.com/Nandaka/PixivUtil2/issues)?

### Description

Each time the program is launched and for each image downloaded, the line is prefixed with `ixivDownloader 20210421` in the console.

### Steps to Reproduce

1. Launch the script
2. The first printf of `header()` will have the prefix `ixivDownloader 20210421`
3. Each Member ID page processed will also have this invalid prefix `ixivDownloader 20210421 [1 of 1] MemberId: 988539 Page: 1Member Url: https://www.pixiv.net/ajax/user/988539/profile/all ` (There is also improper spacing here between "Page" and "Member Url")
4. Each Image ID processed which is not already downloaded will have this invalid prefix `ixivDownloader 20210421 [1 of 1] MemberId: 474309 Page: 1 Post 2+0 of 460 ImageId: 89236210Title: あー？何上から目線で見てんだタココラッ！` (More improper spacing between "ImageId" and "Title")

**Expected behavior:**
The name of the program and version would not be displayed in all these circumstances, much less with a typo.

**Actual behavior:**
An incorrect name of the program as well as the version of the program is printed extremely frequently.

### Versions
PixivDownloader2 version 20210421 (if it wasn't obvious already)

Nandaka:
> header

I don't use linux, but ensure your consoles support ANSI escape sequence as it uses [colorama](https://pypi.org/project/colorama/)

https://github.com/Nandaka/PixivUtil2/blob/master/PixivUtil2.py#L91-L97

> console title

Same case here, maybe you need to play around here? Most likely same issue with #948
https://github.com/Nandaka/PixivUtil2/blob/a9f6e4545ba70a528f0c1aeb8c53f52731042f83/PixivHelper.py#L348

cglmrfreeman:
> I don't use linux, but ensure your consoles support ANSI escape sequence as it uses colorama

[...]

PR #950, opened 2021-05-12
fixed regression in console title: using XTerm OSC

whinette:


whinette:
Fixing issues as described in the previous PR:
https://github.com/Nandaka/PixivUtil2/pull/944

Also fixes https://github.com/Nandaka/PixivUtil2/issues/948
Should fix https://github.com/Nandaka/PixivUtil2/issues/947 but have no full fledged linux env right now. Works in WSL

Issue #949, opened 2021-05-10
MEGA link @Fanbox


ProfessorDocJP:
Trying to download posts that has mega links in it. Is there any way to do that?

cglmrfreeman:
I supported a Fanbox member that would only post thumbnails and links to their google drive for the actual content. They would then remove the google drive links after a certain timeframe in their fanbox to "be fair" to the fans who have been supporting them longer. It's essentially content gatekeeping. I am guessing that's the case with your mega links?

Nandaka:
@bluerthanever got any idea? I don't really subscribe fanbox, so I don't know

ProfessorDocJP:
No it is literally a link that they add to the end of the post but never remove. There is a preview of what the image is like and then they have the mega link for the full image and of course it would take forever to go to each mega link for each image.

Nandaka:
no plan for that, but you can create pull request.

ProfessorDocJP:
Thank you.

bluerthanever:
hmm... as far as i know, mega.nz seems to be quite complicated. files from drive.google.com is easy, but not folders.
usually i just go to fanbox and check for new posts, then download with PixivUtil2. so if there are links i would download them manually.
sorry i didn't use my laptop yesterday.

Nandaka:
maybe just trigger an external application and pass the url name? e.g. call jdownloader when it found a link?
No need to create specific handler directly in the script.

bluerthanever:
i found this but it doesn't seem to be working here...
https://jdownloader.org/knowledge/wiki/terminal/index
maybe it requires java?

Nandaka:
yeah, jDownloader uses java. I'm just using that for example.

No need to use that specifically if you can trigger it from cli.

bluerthanever:
hmm... i found this.
https://board.jdownloader.org/showthread.php?t=62651
someone recommended using an extension though.

also in another thread, someone recommended using APIs of `MyJDownloader`.
https://my.jdownloader.org/developers/

looks like a lot of work either way....

didn't find anything that uses some simple commands to accomplish this.

NHOrus:
Simpliest way is to add links to url_list.txt, same as with links in pixiv posts.

NHOrus:
And let user know/deal with it themselves

bluerthanever:
> Simpliest way is to add links to url_list.txt, same as with links in pixiv posts.

yeah, that would be the easiest.

Issue #930, opened 2021-04-05
Fanbox covers in original size


Psyho7:
### Description

Amm, what about download covers in original size?
I mean, at now, downloaded covers are resized
But you can find url to original size
For example:

Post
https://www.fanbox.cc/@dsmile/posts/2080664
Resized cover
https://pixiv.pximg.net/c/1200x630_90_a2_g5/fanbox/public/images/post/2080664/cover/LuzFC8NEE9HcZQTAsUEH5zvI.jpeg
Original size
https://pixiv.pximg.net/fanbox/public/images/post/2080664/cover/LuzFC8NEE9HcZQTAsUEH5zvI.jpeg 

Post
https://www.fanbox.cc/@grande/posts/2071859
Resized cover
https://pixiv.pximg.net/c/1200x630_90_a2_g5/fanbox/public/images/post/2071859/cover/hyHgOfZ1tbMCftdGRukHqnHi.jpeg
Original size
https://pixiv.pximg.net/fanbox/public/images/post/2071859/cover/hyHgOfZ1tbMCftdGRukHqnHi.jpeg

URLs to original size are written without "c/1200x630_90_a2_g5/"

Found it by browser extension [Image Max Url](https://github.com/qsniyg/maxurl)
And generated links you can see by [this site](https://qsniyg.github.io/maxurl/)

### Versions

20210326

Nandaka:
updated to download both resized and full size.
Add regex to replace the `c/1200x630_90_a2_g5/` or similar variation, also update the html generator.

PR #951, opened 2021-05-16
partial implementation of #949 and #930 enhancement adjustments

bluerthanever:
1. removed some codes from implementation full sized cover image enhancement
2. partial implementation of #949, to store urls found in content that are not images, and calls `PixivHelper.write_url_in_description` to write urls to file.
3. updated test cases.

### DIFF
--- a/PixivModelFanbox.py
+++ b/PixivModelFanbox.py
@@ -13,7 +13,7 @@ import PixivHelper
 from PixivException import PixivException
 
 _re_fanbox_cover = re.compile(r"c\/.*\/fanbox")
-
+_url_pattern = re.compile("(https?|ftp|file)://[-A-Za-z0-9+&@#/%?=~_|!:,.;]+[-A-Za-z0-9+&@#/%=~_|]")
 
 class FanboxArtist(object):
     artistId = 0
@@ -94,7 +94,6 @@ class FanboxPost(object):
     imageId = 0
     imageTitle = ""
     coverImageUrl = ""
-    fullCoverSizeUrl = ""
     worksDate = ""
     worksDateDateTime = None
     updatedDate = ""
@@ -125,6 +124,9 @@ class FanboxPost(object):
 
     embeddedFiles = None
     provider = None
+    # 949
+    descriptionUrlList = None
+
 
     def __init__(self, post_id, parent, page, tzInfo=None):
         self.images = list()
@@ -132,7 +134,8 @@ class FanboxPost(object):
         self.imageId = int(post_id)
         self.parent = parent
         self._tzInfo = tzInfo
-
+        # 949
+        self.descriptionUrlList = list()
         self.linkToFile = dict()
 
         self.parsePost(page)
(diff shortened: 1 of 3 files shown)


bluerthanever:
i changed some codes from 930 implementation because i think some are not so necessary, for example the new field for `FanboxPost` which in codes is actually set the same as `coverImageUrl`, or you could say the other way around...
also before the implementation, cover images would be added to embedded files, but you seem to have removed that. so i added that back...

PixivModelFanbox.py
```diff
@@ -159,14 +162,11 @@ def __str__(self):
...
-        self.coverImageUrl = jsPost["coverImageUrl"]
+        coverUrl = jsPost["coverImageUrl"]
         # Issue #930
-        if self.coverImageUrl is not None:
-            original_url = jsPost["coverImageUrl"]
-            self.fullCoverSizeUrl = _re_fanbox_cover.sub("fanbox", original_url)
-            self.coverImageUrl = self.fullCoverSizeUrl
-            # self.try_add(self.fullCoverSizeUrl, self.images)
-            # self.try_add(self.fullCoverSizeUrl, self.embeddedFiles)
+        if not self.coverImageUrl and coverUrl:
```
Nandaka, line 167: If the self.coverImageUrl already set, then it won't set the self.fullCoverSizeUrl anymore.

PixivModelFanbox.py
```diff
@@ -434,11 +450,7 @@ def WriteHtml(self, html_template, useAbsolutePaths, filename):
 
         cover_image = ""
         if self.coverImageUrl:
-            # Issue #930
-            if self.fullCoverSizeUrl:
-                cover_image = f'<div class="cover"><a href="{self.fullCoverSizeUrl}" target="_blank"><img style="max-width: 1200px; height:auto;" src="{self.coverImageUrl}"/></a></div>'
-            else:
-                cover_image = f'<div class="cover"><img src="{self.coverImageUrl}"/></div>'
+            cover_image = f'<div class="cover"><img src="{self.coverImageUrl}"/></div>'
```
Nandaka, line 453: need to adjust the cover size, else it will be too big in the html. That is why I add inline style here.

bluerthanever: i tested with default html template, and `%coverImage%` is placed under a `<div class="main">` which is under a `<div class="root">`, and in css there are these parts: `.root{max-width:1280px; margin:auto; background-color:#ffffff;}` and `*{margin:0px; padding:0px; max-width:100%; overflow:auto;}`, so the max-width is set to 1280px....
unless the user uses a custom template, in which case the user could modify the style by setting the styles of class "cover" for this `div` and tokens under it.
putting the max-width in the token made it impossible to change it without changing the codes, so i removed it.

also the `href` inside `a` token and the `src` inside `img` token are actually set the same, so i just removed the `a` token wrapping outside the `img` token.

that's my opinion though.

PixivModelFanbox.py
bluerthanever: i removed the `fullCoverSizeUrl` field, and set the `coverImageUrl` to the full sized image url inside the if. =)

PixivModelFanbox.py
Nandaka: gotcha

PixivModelFanbox.py
Nandaka: ok, looks like `coverImageUrl` only set here, so there is no risk that it is already have some values and the big cover is not resolved

PixivModelFanbox.py
bluerthanever: glad we are on the same page
;P

PR #963, opened 2021-06-05
#951 patch

bluerthanever:
the url regular expression uses braces, thus when using `re.findall`, the results are what are in the first braces...
changed the codes to use `re.finditer` to get full url string.

### DIFF
--- a/PixivModelFanbox.py
+++ b/PixivModelFanbox.py
@@ -334,9 +334,9 @@ class FanboxPost(object):
                         # links = re.finditer("<a.*?href=(?P<mark>'|\")(.+?)(?P=mark)", embedStr)
                         # for link in links:
                         #     self.try_add(link.group(2), self.descriptionUrlList)
-                        links = _url_pattern.findall(embedStr)
+                        links = _url_pattern.finditer(embedStr)
                         for link in links:
-                            self.try_add(link, self.descriptionUrlList)
+                            self.try_add(link.group(), self.descriptionUrlList)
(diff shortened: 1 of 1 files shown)


bluerthanever:
i meant brackets

Nandaka:
I'm not sure what is the differences here? from https://stackoverflow.com/a/3766824, it should return the same result.

bluerthanever:
```
>>> import re
>>> url_pattern = re.compile("(https?|ftp|file)://[-A-Za-z0-9+&@#/%?=~_|!:,.;]+[-A-Za-z0-9+&@#/%=~_|]")
>>> text = "https://www.pixiv.net/fanbox/creator/66210"
>>> for item in url_pattern.findall(text):
...     print(item)
...
https
```

this is my test result. and as a result, the url added to url_list is just "https"

or maybe it's not because of the brackets?

bluerthanever:
```
>>> url_pattern.findall(text)
['https']
```

Nandaka:
same result with finditer
image

bluerthanever:
```
>>> for item in url_pattern.finditer(text):
...     print(item.group())
...
https://www.pixiv.net/fanbox/creator/66210
>>> for item in url_pattern.finditer(text):
...     print(item.groups())
...
('https',)
```
i guess it's because of `groups()`?

Nandaka:
ahh, got it. misread group() and groups()

bluerthanever:
=P

Issue #1037, opened 2021-11-01
Indicate the default value of the options y/n, a/i/m


RubenGarcia:
### Prerequisites

* [Y ] Did you read FAQ section in [readme.md](https://github.com/Nandaka/PixivUtil2/blob/master/readme.md)?
* [Y ] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?
* [ Y] Did you search for existing issues in [Issues](https://github.com/Nandaka/PixivUtil2/issues)?

### Description

In 
the various variations of 
    input('Include Pixiv Sketch [y/n]? ').rstrip("\r") or 'n'
do not indicate which is the default option to the user.

I'd like to change all these to either
```
    input('Include Pixiv Sketch [y/N]? ').rstrip("\r") or 'n'
```
or
```
   input('Include Pixiv Sketch [y/n, default n]? ').rstrip("\r") or 'n'
```
Which option do you prefer? I'll create a merge request once this is decided.
### Steps to Reproduce

1. Press 1
2. Answer the questions from the program.

**Expected behavior:** [What you expected to happen]
Default behaviour shown to the user.
**Actual behavior:** [What actually happened]
Default behaviour hidden from the user
**Log file:** [Attach the `pixivutil.log` file in the application folder, recommended to delete the old file, reproduce the issue, and upload the newly generated file here]
N/A
### Versions
v202111031-beta3

Nandaka:
`input('Include Pixiv Sketch [y/n, default is no]? ').rstrip("\r") or 'n'` is preferable.

RubenGarcia:
https://github.com/Nandaka/PixivUtil2/pull/1039

Nandaka:
merged

PR #1039, opened 2021-11-02
[1037] Indicate default value for options

RubenGarcia:


RubenGarcia:
Closes https://github.com/Nandaka/PixivUtil2/issues/1037

PR #1052, opened 2021-11-18
Some optimizations

bluerthanever:
For details please refer to each commits.

### DIFF
--- a/PixivConfig.py
+++ b/PixivConfig.py
@@ -219,12 +219,12 @@ class PixivConfig():
         value = getattr(self, "proxyAddress", None)
         if not value:
             return None
-        match = re.match(r"^(?:(http|socks[45])://)?(\d+\.\d+\.\d+\.\d+):(\d+)$", value)
+        match = re.match(r"^(?:(https?|socks[45])://)?([\w.-]+)(:\d+)?$", value)
         if not match:
             return None
-        scheme, ip, port = match.groups()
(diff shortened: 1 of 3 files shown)


bluerthanever:
I didn't put cases where a user uses proxies with domains instead of ips into consideration...
so if a user uses proxies with domains, the regular expression pattern need to be changes.

bluerthanever:
I didn't put cases where a user uses proxies with domains instead of ips into consideration...
so if a user uses proxies with domains, the regular expression pattern needs to be changed.

PixivConfig.py
```diff
@@ -206,12 +210,23 @@ class PixivConfig():
...
         for item in self.__items:
             setattr(self, item.option, item.process_value(item.default))
-        self.proxy = {'http': self.proxyAddress, 'https': self.proxyAddress}
+
+    @property
+    def proxy(self):
+        value = getattr(self, "proxyAddress", None)
+        if not value:
+            return None
+        match = re.match(r"^(?:(http|socks[45])://)?(\d+\.\d+\.\d+\.\d+):(\d+)$", value)
```
Nandaka, line 222: maybe can use this pattern `^(https|http|socks[45])://[\w.]+(:?\d*)$`
image

Nandaka: updated to allow `-` = `^(https|http|socks[45])://[\w.-]+(:?\d*)$`
image

PR #1055, opened 2021-11-21
Replace lxml with html.parser

Baa14453:
I didn't notice that lxml required an external dependency.
Rather than add it, I've just replaced it with the native html.parser, apparently it's slightly slower but shouldn't matter for this.

### DIFF
--- a/PixivImage.py
+++ b/PixivImage.py
@@ -258,7 +258,7 @@ class PixivImage (object):
 
         # Strip HTML tags from caption once they have been collected by the above statement.
         if self.stripHTMLTagsFromCaption:
-            self.imageCaption = BeautifulSoup(self.imageCaption, "lxml").text
+            self.imageCaption = BeautifulSoup(self.imageCaption, "html.parser").text
 
     def ParseUgoira(self, page):
         # preserve the order
--- a/PixivImage.py
+++ b/PixivImage.py
@@ -258,7 +258,12 @@ class PixivImage (object):
 
         # Strip HTML tags from caption once they have been collected by the above statement.
         if self.stripHTMLTagsFromCaption:
-            self.imageCaption = BeautifulSoup(self.imageCaption, "html.parser").text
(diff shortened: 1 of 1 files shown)


Nandaka:
ah, it is was old code, should have used `"html5lib"` instead like below code.

https://github.com/Nandaka/PixivUtil2/blob/master/PixivBrowserFactory.py#L264

Baa14453:
Ok switched to html5lib instead.

I've also added some extra steps to try and preserve some of the text data.

#### Without Stripping Tags:
```
[初音ミクシンフォニー2021]公式パンフレットにて鏡音リン・レンのイラストを描かせていただきました。&lt;br /&gt;改めましてKAITOさん15周年おめでとうございます！&lt;br /&gt;&lt;a href="/jump.php?https%3A%2F%2Fsp.wmg.jp%2Fmikusymphony%2F" target="_blank"&gt;https://sp.wmg.jp/mikusymphony/&lt;/a&gt;
```
image

#### Original:
```
[初音ミクシンフォニー2021]公式パンフレットにて鏡音リン・レンのイラストを描かせていただきました。改めましてKAITOさん15周年おめでとうございます！https://sp.wmg.jp/mikusymphony/
```
image

#### New:
```
[初音ミクシンフォニー2021]公式パンフレットにて鏡音リン・レンのイラストを描かせていただきました。 改めましてKAITOさん15周年おめでとうございます！https://sp.wmg.jp/mikusymphony/ (https://sp.wmg.jp/mikusymphony/)
```
image

It's not a great example because in this case the HREF tag and <a> contents were the same, but it helps preserve data in cases where its not such as <a href="https://important.site">Click Here to go to my site<a>.
Original:
`Click Here to go to my site`

New:
`Click Here to go to my site (https://important.site)`

Baa14453:
Wait this won't work if there's more than one link...
I will do more testing and update later this week...

Issue #1073, opened 2022-01-19
[Feature/Enhancement] Make an optional "Download by member_id" sketch choice in config.ini


cglmrfreeman:
### Prerequisites

* [X] Did you read FAQ section in [readme.md](https://github.com/Nandaka/PixivUtil2/blob/master/readme.md)?
* [X] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?
* [X] Did you search for existing issues in [Issues](https://github.com/Nandaka/PixivUtil2/issues)?

### Description

Having now just discovered that the default of downloading a member's sketches as "No" since #1037 was implemented, there should be a config which, if set, can replace this dialog. For casual users, they don't really need the option. For archivists, I would very much like the option. Something like `alwaysDownloadSketches = True` or something in config.ini.

~~Additionally, unlike the filenames of the Pixiv Illustration/Manga IDs, the actual URL of the filename of the sketch is rather useless.~~

~~For example, sketch's URL can be `https://sketch.pixiv.net/items/7281939573486935017` while it's hosted URL would be `https://img-sketch.pixiv.net/uploads/medium/file/1810455/3367949766029345023.png`. Currently PixivUtil2 defaults to save `3367949766029345023` of the filename, which is completely useless. If the id of the item, `7281939573486935017`, is saved, the url of the original work can be easily accessed.~~

EDIT: Found out `%image_id%` acts as the "Post ID" for sketches.

### Steps to Reproduce

1. Download by Member Id (option 1)
2. The prompt is always asked and there is no config option to skip this step.
~~3. Sketches are downloaded with IDs that do not relate to accessible web URLs~~

**Expected behavior:**
The prompt for downloading sketches is skipped because it's configured in the `config.ini` file to automatically mark the desired option~~, and sketch works download with a parse-able ID to reverse navigate to the original URL the work was obtained from~~.

**Actual behavior:**
The prompt to download sketches is always displayed and defaults to `no`~~, and sketch files are downloaded with an ID that is impossible to reverse navigate to~~.

**Log file:**
None necessary

### Versions
PixivDownloader2 version 20211104

cglmrfreeman:
It looks like PixivUtil2 already grabs the "Post ID"? Can that be used in the filename?

cglmrfreeman:
[...]

PR #1078, opened 2022-02-02
Pixiv Sketch config and Database Manager changes

cglmrfreeman:
I've been looking at the code for the database manager as well as how sketches are handled to create this pull request, and I've made a few changes that I hope work to improve the codebase as a whole:

- If a Sketch was found in the database and overwrite and alwaysCheckFileSize were disabled, the sketch would not be skipped. The message informing the user that it was skipping the post also included an erroneous " ." after the sentence. These have been fixed.
- The interactive cleanup for Fanbox had the incorrect index to display the `save_name` from the sqlite query, it would display the `page` value instead. This has been fixed.
- The menu interface for database management used the old implementation prior to the inclusion of the `colorama` requirement. It has been updated to more closely resemble the main PixivUtil2 menu.
- Clarified in the `readme.md` that the `%image_id%` filenameFormat for sketch posts would use the "Post id".
- A KeyboardInterrupt in PixivDownloadHandler would throw a `PIXIVUTIL_ABORTED` but a `PIXIVUTIL_KEYBOARD_INTERRUPT` is already defined. I feel it would be better to distinct an abort caused by software error versus a user initiated interrupt in this instance.
- Pixiv sketches had no cleanup options. They do now
- The variable `defaultSketchOption` has been defined in the configuration to skip the "Include Pixiv Sketch" option when downloading by `member_id` as requested in  #1073. Just like the option used currently, if set to `y` it will include sketches, if set to `n` it will exclude sketches. If the value is not set or set to an invalid option, it will continue to prompt the user to include sketches or not. The `readme.md` has also been updated to include this config change.

Please let me know what you think. Thank you.

EDIT: What this does _not_ fix is handling a keyboard interrupt while PixivUtil2 downloads sketches. Unlike the regular `member_id` illustration download, a keyboard interrupt does not pause downloading, and simply moves onto the next sketch. You can not fully abort a sketch download like you can with a normal illustration download.


cglmrfreeman:
Hmm, it appears that with the `PIXIVUTIL_ABORTED` changed to `PIXIVUTIL_KEYBOARD_INTERRUPT`, this makes the `PIXIVUTIL_ABORTED` constant unused. Functionally the program appears to be no different swapping the two, but "ABORTED" may have been chosen on purpose for reasons I can't find clear at the moment.

cglmrfreeman:
I have merged upstream and resolved conflicts, but github is saying conflicts exist and I have not merged upstream. Any ideas, @Nandaka?

cglmrfreeman:
I don't understand why now it's showing your commit changes in my changes here? They should be caught up with master?

cglmrfreeman:
Any thoughts so far? Questions? Comments? Concerns?

Nandaka:
>  Functionally the program appears to be no different swapping the two, but "ABORTED" may have been chosen on purpose for reasons I can't find clear at the moment.

I think it was used before as generic exception.

Looks ok 😄

cglmrfreeman:
LOL! I wasn't finished but sure! I guess this works ahaha! :sweat_smile:

Nandaka:
welp, I thought u waiting to change for me to change the draft to reviewed :P

cglmrfreeman:
@Nandaka LOL! No worries, the only thing I had left to do was allow the cancel prompt when downloading Pixiv sketches, which can be handled in another PR. Currently keyboard interrupt cancels the currently downloading sketch and moves onto the next.

Nandaka:
yeah, I prefer separate small PR, rather than big changes in single PR, else it is difficult to follow :P

PR #1104, opened 2022-03-19
add `demjson3 ` into requirements.txt

DenDen047:


cglmrfreeman:
Why? What requires this?

Nandaka:
it is should be in the requirement already...

PR #1105, opened 2022-03-19
Update Dockerfile to install ffmpeg

DenDen047:
add ffmpeg installation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,11 @@
 FROM python:3.8
 
-COPY requirements.txt requirements.txt
+RUN apt-get update
 
+COPY requirements.txt requirements.txt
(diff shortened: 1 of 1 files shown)


cglmrfreeman:
Should it not be 
```
RUN apt-get update && apt-get install -y \
     ffmpeg
```
?

Nandaka:
based on limited knowledge of *nix, isn't it `apt-get update` will update all installed apps/libs in the system/docker image?

DenDen047:
> based on limited knowledge of *nix, isn't it `apt-get update` will update all installed apps/libs in the system/docker image?

No. `apt-get update` only update the package list. `apt-get upgrade` will update all the installed apps and libs.

Issue #1109, opened 2022-04-02
[Bug] jpeg to gif conversion issue


Toyem:
### Prerequisites

* [x] Did you read FAQ section in [readme.md](https://github.com/Nandaka/PixivUtil2/blob/master/readme.md)?
* [x] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?
* [x] Did you search for existing issues in [Issues](https://github.com/Nandaka/PixivUtil2/issues)?

### Description

I wanted to use the option to convert ugoira to gif. 
It turns out that this feature works quite well except for some ugoira from users 4734113.
We notice that the gif obtained truncates a large number of images and keeps only a few images at the end.

### Steps to Reproduce

To do this, change the option `createGif` to True in the category [Ugoira] in the config.ini file.
Then launch `PixivUtils.py` by selecting option 2.  Download by image_id and enter one of the following two examples (sorry for the not family firendly character of the chosen examples, it's the only ones that don't work):
- 91281700
- 96990356

### Small investigation

When we display the logs in debug mode of the command passed in the function `convert_ugoira` of `PixivHelper.py` by the line `subprocess.Popen`, we realize that a great number of automatic conversion of pixel_format are realized and that at certain moments, EOI tags seem to be missing.
```
[mjpeg @ 0x7ff293c6bd00] EOI missing
```
```
[auto_scale_1 @ 0x7fc817d18300] w:300 h:300 fmt:gray sar:1/1 -> w:300 h:300 fmt:bgra sar:1/1 flags:0x0
[auto_scale_1 @ 0x7fc81980c380] w:300 h:300 fmt:yuvj444p sar:1/1 -> w:300 h:300 fmt:bgra sar:1/1 flags:0x0
```
After inspection, the issue does not come from the images (jpeg) which have all their SOI and EOI tags (respectively "\xff\xd8" and "\xff\xd9"). I put the mini script I used in footnote if you want to check the info.

However, the problem seems to come from the automatic conversion of the images which cuts the generation of the gif and erases the information previously generated.
To realize this, you just have to delete the images generating an automatic conversion (images 21, 22, 23, 106 and 107 for the artwork id 91281700) from the file `i.ffconcat` to prevent their use during the conversion and the gif is generated from the image 00000.jpeg without issues.

[...]

PR #1110, opened 2022-04-03
Fix bug #1109

Toyem:
Added a workaround that performs a bit depth check for each image contained in the temporary ugoira processing folder. If some images have a lower bit depth number than others, these images are all encoded to have the same bit depth number corresponding to the highest among the images in the folder.

### DIFF
--- a/PixivHelper.py
+++ b/PixivHelper.py
@@ -30,6 +30,7 @@ from typing import Union
 
 import mechanize
 from colorama import Fore, Style
+from PIL import Image, ImageFile
 
 import PixivConstant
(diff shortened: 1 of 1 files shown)


cglmrfreeman:
Now should we add a reprocess function to scan all the `ugoira` manga types and overwrite the old/bad files? :thinking:

Toyem:
t's clearly a good idea but it's not really a bug fix but the implementation of a new feature. 
I have the impression that the id's already downloaded are skipped, so we will have to make sql calls and restart the downloads. The idea is good but for the time being, I don't really have the time

PR #1137, opened 2022-07-21
Fixes Byte Order Mark in Downloaded file list, again.

NHOrus:
codecs.open is stateless, so simple way doesn't work and we need to deal
with state, or we get a BOM on each newline. This is suboptimal and forces
more postprocessing of Downloaded file list for correct result.

All of this code could be deleted if irfan fixes it's unicode file bug.
Alas, bug persists in all text files irfan opens, including it's config.
So fix is unlikely. :^(

Note to self: test not only on one line, but on multiple lines.

### DIFF
--- a/PixivDownloadHandler.py
+++ b/PixivDownloadHandler.py
@@ -245,6 +245,14 @@ def download_image(caller,
                         raise
                 else:
                     PixivHelper.print_and_log('info', ' done.')
+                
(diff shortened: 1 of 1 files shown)


Nandaka:
looks like you missed one place to replace?
image

supposed like this? else it have no effect.
image

Issue #1159, opened 2022-09-09
Download limit when using bookmark count


YukihoAA:
### Prerequisites

* [x] Did you read FAQ section in [readme.md](https://github.com/Nandaka/PixivUtil2/blob/master/readme.md)?
* [x] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?
* [x] Did you search for existing issues in [Issues](https://github.com/Nandaka/PixivUtil2/issues)?

### Description

When using bookmark-count for job_type 3 it will cause IP download limit from pixiv server.

(server returns `None` for search page dump)

i added this code to PixivBrowerFactory.py L892 

`PixivHelper.wait(result, self._config)`

and wait 3 seconds interval for retriving bookmark infomation, it does not works too.

### tested batch_job option

`        "Chino": {
            "enabled": true,
            "job_type": "3",
            "tags": "チノ 香風智乃",
            "start_page": 1,
            "end_page": 0,
            "wildcard": true,
            "title_caption": false,
            "start_date": "2022-06-01",
            "bookmark_count": 300,
			"sort_order": "date_d",
            "type_mode": "a",
            "option": {
                "useTagsAsDir": true
            }
        }
`

### Dump
`None`

PR #1160, opened 2022-09-10
Remove retriving bookmarks function

YukihoAA:
for #1159 

removing retriving bookmarks infomation to prevent ban.

it works fine for tags mode but idk why code get bookmarks seperatly before download...

### DIFF
--- a/PixivBrowserFactory.py
+++ b/PixivBrowserFactory.py
@@ -864,7 +864,7 @@ class PixivBrowser(mechanize.Browser):
                     result.parseTags(response_page, tags, current_page)
 
                     # parse additional information
-                    if use_bookmark_data:
+                    if 0:
                         idx = 0
                         print("Retrieving bookmark information...", end=' ')
                         for image in result.itemList:
@@ -889,6 +889,7 @@ class PixivBrowser(mechanize.Browser):
                                 image_info_js["body"]["bookmarkCount"])
                             image.imageResponse = int(
                                 image_info_js["body"]["responseCount"])
+                            PixivHelper.wait(result, self._config)
                     print("")
                 except BaseException:
                     PixivHelper.dump_html(f"Dump for SearchTags {tags}.html", response_page)
--- a/PixivTagsHandler.py
+++ b/PixivTagsHandler.py
@@ -50,8 +50,8 @@ def process_tags(caller,
         last_image_id = -1
         skipped_count = 0
         use_bookmark_data = False
-        if bookmark_count is not None and bookmark_count > 0:
-            use_bookmark_data = True
+        #if bookmark_count is not None and bookmark_count > 0:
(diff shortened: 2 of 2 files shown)


Nandaka:
It is to skip the post if the bookmark limit is defined when you download by tags. If you remove this code, then it will download all images when you set a limit.

If you want to skip this bookmark data retrieval, then just set the limit to 0/None from the console input.

https://github.com/Nandaka/PixivUtil2/blob/master/PixivUtil2.py#L366
https://github.com/Nandaka/PixivUtil2/blob/master/PixivUtil2.py#L371

YukihoAA:
@Nandaka 

bookmark skipping and grouping works fine for image mode because there is skipping function in PixivImageHandler.py L174

image

image

Nandaka:
But if the image doesn't exists in the db, it still need to get the bookmark data. Only if you are registered premium user that have the data from the json response, I think?

https://github.com/Nandaka/PixivUtil2/blob/b723feb4a72567746ab8696586fb3647a5d49334/PixivBrowserFactory.py#L832

fyi, bookmark data is empty from server
image

and it is only available when you load the page, which I think it is more expensive to run? and anyway you still need to get the page anyway if you want to filter by bookmark count.
image

Maybe you can add a logic to skip this you are premium user? e.g. `if use_bookmark_data and not self._isPremium:`

https://github.com/Nandaka/PixivUtil2/blob/b723feb4a72567746ab8696586fb3647a5d49334/PixivBrowserFactory.py#L867

YukihoAA:
I'm not using premium now and I'm using bookmark count skip feature.

It was an ideal codes **to use bookmark count skipping with out being IP ban.** issue #1159 
if i use 20220825 version, like if image skipped 120 in one time, pixiv will give me IP ban.

I tried to add 2 second wait to every loop for bookmark retriving function using wait(result, self.__config) 
but does not worked :(

so my Idea was just using functions in PixivImageHandler to make some delay to prevent ban.

Nandaka:
can you confirm after running it for some time? because after the change, then it will always fetch the image page and then skipping it if the bookmark count is lower ==> which in turn, currently it will cause 429 error due to it request too many image page too often.

YukihoAA:
In case v20220825: will got baned after 120~180 image skipped by retriving bookmarks with 5 sec delay
image

After edited: it worked fine (~2hr running) with 5 sec delay.
image
image

I runned same option that already downloaded so almost all images was skipped.


2 sec delay is not working both case.

YukihoAA:
3sec is seems fine too
image
image
image

YukihoAA:
3sec is seems fine too (1hr test) - @Nandaka 
image
image
image
image
image

Nandaka:
sorry for the late reply, I was just back. Anyway, it is merged now

Issue #796, opened 2020-09-16
Error when creating png file of a particular image.


polujm9:
### Prerequisites

* [x] Did you read FAQ section?
* [x] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?

### Description

I'm downloading images as usual and a problem appeared when downloading particular image (stop when creating apng file).

I had try to upgrade PixivUtil2 to v20200915 and re-download ffmpeg.exe and not working.

### Steps to Reproduce

1. download image id = 72778077

**Expected behavior:** Create files successfully

**Actual behavior:** Stopped when creating png file

### Versions

Logs in pixivutil.log :
```
2020-09-16 13:48:42,476 - PixivUtil20200915 - INFO - ###############################################################
2020-09-16 13:48:42,476 - PixivUtil20200915 - INFO - Starting with argument: [E:\Downloads\PixivUtil2.exe].
2020-09-16 13:48:42,480 - PixivUtil20200915 - INFO - Setting log level to: DEBUG
2020-09-16 13:48:42,480 - PixivUtil20200915 - INFO - No default cookie jar available, creating... 
2020-09-16 13:48:42,988 - PixivUtil20200915 - INFO - Using default DB Path: E:\Downloads\db.sqlite
2020-09-16 13:48:42,994 - PixivUtil20200915 - INFO - Only process members where the last update is >= 7 days ago
2020-09-16 13:48:43,855 - PixivUtil20200915 - INFO - Using Username: zxczxc123456789
2020-09-16 13:48:43,855 - PixivUtil20200915 - INFO - Trying to log in with saved cookie
2020-09-16 13:48:44,126 - PixivUtil20200915 - INFO - Logging in, return url: https://www.pixiv.net/
2020-09-16 13:48:44,126 - PixivUtil20200915 - INFO - Login successful.
2020-09-16 13:48:44,127 - PixivUtil20200915 - INFO - Logged in using cookie
2020-09-16 13:48:44,128 - PixivUtil20200915 - INFO - My User Id: 3753063.
2020-09-16 13:48:44,129 - PixivUtil20200915 - INFO - Premium User: False.
2020-09-16 13:48:44,129 - PixivUtil20200915 - INFO - Locale = 
2020-09-16 13:48:50,811 - PixivUtil20200915 - INFO - Tags mode (3).
[...]

PR #1179, opened 2022-10-10
Tell ffmpeg not to reset filter state for ugoira encode

hdk5:
Related to #1109 #1110

There's no need to explicitly re-encode each frame, as `ffmpeg`'s option `reinit_filter` allows to ignore input frames' parameters inconsistency. 

More info on Super User: https://superuser.com/questions/1465039/.

This PR removes `check_image_encoding` function and adds `-reinit_filter 0` to `ffmpeg`'s params.

Tested by gif encoding illustrations id 101814634 (which broke my own personal download scripts) and 91281700, 96990356 (from #1109)


Nandaka:
@Toyem ping for comment?

Toyem:
I can check the commit tonight on the 3 major OS with a bunch of images I know are problematic but it looks much cleaner than what I coded. 
If everything works, it's a great improvement 👍

Toyem:
@Nandaka I've tested it with Linux (Ubuntu 22.04) and MacOS Catalina 10.15.7 and it's working perfectly fine, much faster and with the same result.
I couldn't test it with Windows cause my hard drive decided to corrupt himself tonight but for what i can test, it's a perfectly fine improvement.

Toyem:
Wait ! i spoke too soon, i got an error when trying to download 101814634 as an apng.
What's weird is that the apng file is here and perfectly fine but i still have an error, is it just me ?
[pixivutil.log](https://github.com/Nandaka/PixivUtil2/files/9760207/pixivutil.log)

hdk5:
Sorry, my bad - I did not actually check for other output formats conversion.

It seems `hqdn3d` filter for `apng` conversion doesn't work properly with `-reinit_filter 0`.
I'll look into what I can do with this a bit later.

hdk5:
I found that the usage of `hqdn3d` filter was explained in https://github.com/Nandaka/PixivUtil2/issues/796#issuecomment-693232061, but I couldn't reproduce mentioned there "not all frame will be processed" problem - not with direct `ffmpeg` invocation at least. Is it still relevant? For 101814634, it also increases output apng size in almost three times.

Anyways, it seems that empty `scale` filter as the first one in the list just does the required job of converting all of the input frames into uniform pixel format to be used by the following filters.

Also, an irrelevant thing I found while testing 72165291 encode (from https://github.com/Nandaka/PixivUtil2/issues/796#issuecomment-693232061): ffmpeg may drop frames with the current encoding params (input 150 -> output 114).
Calculating correct input/output framerate or just dumbly passing `-r 999` for output would probably fix it.

Toyem:
I tested it and it seems to work (still not tested on Windows, did not have time to reinstall the whole VM)

NewUserHa:
how could images from a zip have different color formats in https://github.com/Nandaka/PixivUtil2/issues/1109?
I think it should avoid re-encoding to prevent quality loss.

Issue #439, opened 2018-11-16
[Feature Request] Have options to download both R-18/18G or one over other.


ghost:
I was wondering if there would be an option or confirmation to download R-18G Files alongside the regular R-18 Only Mode like if you were in said mode and were going to download some stuff, the console would ask if you wanted to download both kinds or one over the other.

The R-18 mode only downloads things that are tagged with said mode and not the other variants of said mode.

Nandaka:
Currently not possible for member related images, as the filter from Pixiv only allow to supply 1 tag (either R-18 or R-18G).
What I can do is to run 2 query for each tag. Please note that if the images has both tag, it will be processed twice.

LatvijaSuido:
Thank you for the informative knowledge. I was wondering this myself. It should not be challenging for people to open up the .ini and change r18mode to True if they wish to get a full catalogue from a member or section. This seems like a low priority goal. I may be wrong and minority. English is not first language after all :)

Nandaka:
hmm, I can set it in the `config.ini` as flag `r18Type` with value `0` = both R18 and R-18G, `1` = only R18, and `2` = only R18G

Issue #420, opened 2018-10-03
the timestamp is delayed by 9 hours.


nalvoks:
### Prerequisites

* [x] Did you read FAQ section?
* [ ] Did you test with the latest releases or commit?
not yet.
I'm using 20180918.
I'll try after download is completed.

### Description

the timestamp of downloaded file is delayed by 9 hours.
I'm in Japan (UTC+9) and I think that's the reason.

### Steps to Reproduce

using PixivUtil2 with --startaction=4

Nandaka:
It is stored in utc.
for example, https://www.pixiv.net/ajax/illust/69925083
https://github.com/Nandaka/PixivUtil2/blob/master/PixivModelWhiteCube.py#L234

I'm not sure if I should convert to local pc time? maybe can use something like https://stackoverflow.com/a/10854983

nalvoks:
oh ok.
no problem if it's the intended behavior.

thanks.

Nandaka:
I think in the web ui, it will be adjusted to the time zone based on your location in the profile settings?

For example:
https://www.pixiv.net/member_illust.php?mode=medium&illust_id=70995384
Web shows October 3, 2018 6:09 AM
![image](https://user-images.githubusercontent.com/836867/46511590-1ba67e80-c882-11e8-8370-30197abd7778.png)
API shows 2018-10-02T22:09:47+00:00 (9h differences)
![image](https://user-images.githubusercontent.com/836867/46511636-44c70f00-c882-11e8-98cb-3cbe8171d2fe.png)

What is the date shown in your side (assuming your location is not set to tokyo/japan)?

nalvoks:
the date shown in my side is

https://www.pixiv.net/ajax/illust/70850475
"createDate":"2018-09-24T09:36:51+00:00"

https://www.pixiv.net/member_illust.php?mode=medium&illust_id=70850475
September 24, 2018 18:36

downloaded file
20180924 09:36:51

nalvoks:
i forgot this.

70850475.txt
Date          = 09/24/18 09:36

Nandaka:
I assume, you also setting the location to tokyo/japan? if not, then meaning the timestamp in the web page is always use japan as the time zone offset.

nalvoks:
Yes. the location is set to Japan. and the system locale is also Japan.
but the timestamp of downloaded file is in UTC+00:00 time zone.

Nandaka:
[...]

PR #1197, opened 2022-11-19
Enhance document

NewUserHa:
This PR tries to update the document and add some details to it.

Found some issues that need to discuss:
* `customCleanUpRe` no document.
* `customBadChars` no code no use.
* `writeRawJSON` no code no use.
* `writeSeriesJSON` no code no use. the `includeSeriesJSON` do the job.
* `useLocalTimezone` can be removed since all file time is always use local time.

hope this can make the document better for coming new users


cglmrfreeman:
What on earth?

NewUserHa:
the document, the readme.md.

those settings are not used at all or having issues.

Nandaka:
Is the document you are referring to is the Readme.md?
 
`writeRawJSON` it is used in the code and in the config file.

image
image

`includeSeriesJSON` is the correct one, not `writeSeriesJSON` ==> only need to update the readme

`useLocalTimezone` => not in my case see #420 as the API return in utc

`customCleanUpRe` is used as part of filename sanitizer if user want to use it.
image

`customBadChars` ==> I think already replaced with above, at least I cant find the usage beside in the config.

NewUserHa:
I'm trying set config based on the document.

> writeRawJSON, writeImageJSON

turned both on, but no '..-raw.text' file written, only a json.text from 'writeImageJSON'

> useLocalTimezone

system use time in seconds since the epoch (i.e. UTC) always.
the reason of '70850475.txt' has a wrong file time may be no 'os.utime()' followed after '.close()' at https://github.com/Nandaka/PixivUtil2/blob/f2e9ab19fe471c93c0a1697d89f1f9f93f3c39cd/PixivImage.py#L434
tzinfo is useful in string format and time calculation
so useLocalTimezone may be no need

> customCleanUpRe, customBadChars

I think maybe adding a few example will help

NewUserHa:
some new useful info that I think can help users :
> createWebp, webm, apng, gif

https://github.com/Nandaka/PixivUtil2/blob/f2e9ab19fe471c93c0a1697d89f1f9f93f3c39cd/PixivHelper.py#L1045
this line extracted all files from the ugoira. it can directly convert from the source file .zip to all other formats actually, I guess.

> webpparam: Parameter to be used to encode webm. default is lossless 0 -q:v 90 -loop 0 -vsync 2 -r 999

"to encode webm" ==> "to encode webp".
while actual argument miss lossless 1(true) at https://github.com/Nandaka/PixivUtil2/blob/f2e9ab19fe471c93c0a1697d89f1f9f93f3c39cd/PixivHelper.py#L1014

> webpparam

webp can use lossless for the best quality by default.

checkLastModified: when found db record and both `overwrite` and `checksize` is false, compare the local file's last-modified time with the uploaded time on pixiv.
https://github.com/Nandaka/PixivUtil2/blob/269912590aeaa3652316bb4bd344147df2995fb0/PixivDownloadHandler.py#L75
it should do the time check after finding the file size is the same.

alwaysCheckFileSize: when found db record, if both this and overwrite are false, it will only check if the file exists at the recorded path. otherwise, it'll always do a file size check.

overwrite: need `alwaysCheckFileSize` to work. when `overwrite` is true, if the file size is different, it will just delete it and then redownload it.

readme.md
```diff
@@ -528,7 +528,7 @@ Please refer run with `--help` for latest information.
   Set to `True` to export the image information to a .XMP sidecar file, one per image in the album. The data contained within the file is the same but some software requires matching file names to detect the metadata. If set to `True`, then `writeImageXMP` is ignored.
 - verifyimage
 
-  Do image and zip checking after download. Set the value to `True` to enable.
+  Check if downloaded files is valid image or zip. Set the value to `True` to enable.
```
cglmrfreeman, line 531: If "files" is plural, the proper term to use would be "are", so `Check if downloaded files are a valid image or zip`, however this is not what `verifyimage` actually does. `verifyimage` simply checks if the file ends in ".jpg", ".png", ".gif", ".ugoira", or ".zip". It doe snot verify if it is a valid image or zip file.

PixivConfig.py
```diff
@@ -92,8 +92,8 @@ class PixivConfig():
         ConfigItem("Settings", "downloadAvatar", False),
         ConfigItem("Settings", "useSuppressTags", False),
         ConfigItem("Settings", "tagsLimit", -1),
-        ConfigItem("Settings", "writeImageInfo", False),
         ConfigItem("Settings", "writeImageJSON", False),
+        ConfigItem("Settings", "writeImageInfo", False),
```
cglmrfreeman, line 96: Moving this now separates the two JSON configs from each other. Was there a good reason for this?

readme.md
NewUserHa: mistake.

that adding details is trying to make config PixivUtil2 easier for experienced users.

actually, `verifyimage` will validate files: https://github.com/Nandaka/PixivUtil2/blob/f2e9ab19fe471c93c0a1697d89f1f9f93f3c39cd/PixivDownloadHandler.py#L203-L248

PixivConfig.py
NewUserHa: because `writeImageJSON` is the original data from the api, while `writeImageInfo` is just a few texts extracted from it.

so making the original above others is trying to let users know that the next few settings are just a subset of it.
however, the other settings need some extra info too.

readme.md
cglmrfreeman: Oh so it does, my mistake

NewUserHa: it seems that the `overwrite` and `alwaysCheckFileSize` have too much tech debt.
and not sure how to detail those

Nandaka: haha, it is from 2015 and no major rewrite at all. As long it is working as my expectation, I don't plan to do anything as long it doesn't break.

Nandaka:
also send a lot of changes in one big pull request, i might not be able to understand it 😄

NewUserHa:
This PR will try to improve the document only, to make both new and experienced users clear with some details.

NewUserHa:
The rest of this PR is done. Except the `customCleanUpRe`, but I don't use those settings and didn't read those code.

@Nandaka Can you add its document?

cglmrfreeman:
Did you notice `r18Type` isn't documented? The information on it can be found in #439.

NewUserHa:
@cglmrfreeman it's updated now.
There're many options I don't use. You can also suggest any option I didn't see to the PR directly.

PixivConfig.py
Nandaka: historically, the `writeImageInfo` was come up first, then only followed by `writeImageJSON` when someone ask for the raw json.

PixivConfig.py
```diff
@@ -223,7 +223,7 @@ def proxy(self):
         value = getattr(self, "proxyAddress", None)
         if not value:
             return None
-        match = re.match(r"^(?:(https?|socks[45])://)?([\w.-]+)(:\d+)?$", value)
+        match = re.match(r"^(?:(https?|socks[45]h?)://)?([\w.-]+)(:\d+)?$", value)
```
Nandaka, line 226: should we add check for socks4a too?

PixivConfig.py
NewUserHa: Is the `writeImageInfo` the same as `writeImageJSON` when no `writerawjson`?

PR #1199, opened 2022-11-23
Update to python 3.11 for faster startup

NewUserHa:
Faster startup time can help command line program has a better user experience.
 
detail: https://docs.python.org/3.11/whatsnew/3.11.html#faster-cpython

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: jammy
 language: python
 python:
   - "3.11"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 python:
-  - "3.9"
+  - "3.11"
 # command to install dependencies
 install:
   - pip install -e .
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.8
+FROM python:3.11
 
 COPY requirements.txt requirements.txt
 


cglmrfreeman:
I don't understand. Anyone who doesn't PixivUtil2 in a docker won't benefit from this and the last Travis CI build was [over two years ago](https://travis-ci.org/github/Nandaka/PixivUtil2/builds). If you meant "Update docker file and .travis.yml to Python 3.11", yeah, that will work, but "for faster startup" means nothing. There is no "better user experience" coming from this pull request. The documentation states:
> This has a big impact for short-running programs using Python.

As far as I am aware, PixivUtil2 is not a "short-running" program. Furthermore, to prove this, benchmarks are required. What's the startup time for docker versus your branch?

NewUserHa:
It's just saying that python 3.11 could make console programs startup faster and users will feel happy.

dont see why you use a python file in docker, that way you have to mount the folder anyway.

NewUserHa:
The docker one can speed up the docker of python; The travis ci can run some pip install tests.

It just notices that updating to python 3.11 has benefitted, and this PR is trivial actually. If the next release is bundling python 3.11, can close this PR.

cglmrfreeman:
The next release of PixivUtil2 does not bundle python at all. You can close this PR yourself if you wish, else we're waiting on @Nandaka.

Issue #982, opened 2021-07-13
Trailing . left on image name but not info file.


Baa14453:
### Prerequisites

* [x] Did you read FAQ section in [readme.md](https://github.com/Nandaka/PixivUtil2/blob/master/readme.md)?
* [x] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?
* [x] Did you search for existing issues in [Issues](https://github.com/Nandaka/PixivUtil2/issues)?

### Description

#627 removes the trailing `.` from filenames for info files.
If the title is a `.` this seems to only apply to the downloaded info files and not the image file itself. So they end up with different names.

![image](https://user-images.githubusercontent.com/9057997/125501517-e11f923f-710f-496e-af77-a9cdb5618a57.png)

### Steps to Reproduce

1. Enable Info file downloading
2. Download image `91226078`
3. Observe naming discrepancy.

**Expected behavior:**
Both files with have the same names, different extensions.

**Actual behavior:**
One file ommits the additional `.` the other doesn't.

### Versions

20210702

Nandaka:
Are you using windows?
just for reference for me https://stackoverflow.com/a/16203594

Baa14453:
no, using Ubuntu 20.04
The files create OK but because the filenames are different between the info file and image file, my software doesn't associate them

-------- Original Message --------

Nandaka:
Oo, then it might be a bit difficult for me to test, as windows will automatically truncate the ending period... and I don't put special handling with `\\?\` in windows (i only have windows box)

Baa14453:
I'll see what I can do

PR #1203, opened 2022-12-02
fix: infofile add extension name first

cokemine:
This PR should fix: https://github.com/Nandaka/PixivUtil2/issues/982
I can confirm this issue both happens on Linux and Windows.

The arg of PixivHelper.sanitize_filename() should have a extname for info file otherwise trailing dots and spaces from the title will be removed, which is not as expected.

### DIFF
--- a/PixivImageHandler.py
+++ b/PixivImageHandler.py
@@ -304,8 +304,8 @@ def process_image(caller,
                                                                     searchTags=search_tags,
                                                                     useTranslatedTag=config.useTranslatedTag,
                                                                     tagTranslationLocale=config.tagTranslationLocale)
-                        info_filename = PixivHelper.sanitize_filename(info_filename, target_dir)
-                        image.WriteXMP(info_filename + ".xmp")
+                        info_filename = PixivHelper.sanitize_filename(info_filename + ".xmp", target_dir)
+                        image.WriteXMP(info_filename)
 
             if config.writeImageInfo or config.writeImageJSON or config.writeImageXMP:
                 filename_info_format = config.filenameInfoFormat or config.filenameFormat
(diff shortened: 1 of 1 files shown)


cglmrfreeman:
I'm not too familiar with what's happening in this exact code, but is there a reason [pathlib](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.stem) isn't being used to handle this?

NewUserHa:
indeed should use `PurePath.stem` to trim the last dot in the filename to avoid more issues.

cokemine:
> indeed should use `PurePath.stem` to trim the last dot in the filename to avoid more issues.

I wonder why we have to trim last dots ? All info file should have a ext name `.txt` or `.json`
`....txt` both valid on Windows and Linux

NewUserHa:
> The arg of PixivHelper.sanitize_filename() should have a extname for info file otherwise trailing dots and spaces from the title will be removed, which is not as expected.

Isn't the PR trying to trim the last dot of the filename without the suffix?

I currently didn't encounter any issues with the default filename template. How can `[:-8]` be right for the default filename template?

cokemine:
> > The arg of PixivHelper.sanitize_filename() should have a extname for info file otherwise trailing dots and spaces from the title will be removed, which is not as expected.
> 
> Isn't the PR trying to trim the last dot of the filename without the suffix?
> 
> I currently didn't encounter any issues with the default filename template. How can `[:-8]` be right for the default filename template?

I don't think it's needed to trim the suffix dots if the file has an explicit extension name (.txt or .json)

Before this PR, `.` from image title are trimed. This is because it is passed into `PixivHelper.sanitize_filename()` without the extension.(info file) I added a fake extension `.infoext` to circumvent this, and then write the file with the correct extension.

But for image file, The argument of `PixivHelper.sanitize_filename()` containing the corrent extension name of the image, so `.` won't be trimed


I think `PurePath.stem` has the same effect, but the fake extension I set is a fixed length, so I didn't use it

PR #1204, opened 2022-12-03
fix: remove trim `_pxx`

cokemine:
This should fix for some pictures titled end with _xxx, e.g. https://www.pixiv.net/artworks/100345141

`リコリス・リコイル_02.txt` shouldn't be trimed to `リコリス・リコイル.txt`

However, I don't know why `info_filename = re.sub(r'_p?\d+$', '', info_filename)` code exists. I checked the commit history and found that this was added six years ago, I think it may be a historical problem

### DIFF
--- a/PixivImageHandler.py
+++ b/PixivImageHandler.py
@@ -322,8 +322,6 @@ def process_image(caller,
                                                           searchTags=search_tags,
                                                           useTranslatedTag=config.useTranslatedTag,
                                                           tagTranslationLocale=config.tagTranslationLocale)
-                # trim _pXXX
-                info_filename = re.sub(r'_p?\d+$', '', info_filename)
                 info_filename = PixivHelper.sanitize_filename(info_filename + ".infoext", target_dir)
                 if config.writeImageInfo:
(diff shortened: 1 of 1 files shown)


Nandaka:
it is for manga images, if you don't remove it, the info file will contains _pxx

cokemine:
I 've tested both single image and manga images. I found that deleting this code did not change the output(except the bug that I mentioned), that's why I have that question

cokemine:
I used the default config
```
filenameFormat = %artist% (%member_id%)/%urlFilename% - %title%
filenameMangaFormat = %artist% (%member_id%)/%urlFilename% - %title%
filenameInfoFormat = %artist% (%member_id%)/%urlFilename% - %title%
filenameMangaInfoFormat = %artist% (%member_id%)/%urlFilename% - %title%
```
`info_filename` is end with `%title%`, and `_pxx` will be only trimed from `%title%` not from `%urlFilename%`


So all the manga information will be in the format `id_pXX - title` and `_pXX` won't be trimed

PR #1217, opened 2022-12-28
Add the unused configuration option "extensionFilter" to the app.

Terryble2:
This change allows users to put an extensionFilter configuration in their config.ini.

My personal use case is to only ever download images and never ugoiras with jpg|png|gif.


cglmrfreeman:
So this isn't actually implemented at this point? Just an option has been created for it?

PR #1266, opened 2023-05-01
Add option to manually specify file extension inside file names.

Baa14453:
Digikam recently changed the behaviour of XMP sidecar files and now they need to match the full file-name of the associated image, ie. what used to be:
> image.jpg
> image.xmp

is now:
> image.jpg
> image.jpg.xmp

This PR adds the option to use %file_ext% in the file naming process so I can change my filename config to say "image.%file_ext%".

Now I have to go rename 140,000 sidecar files thanks Digikam...


Nandaka:
you apply this for the `filenameInfoFormat`, right?

Baa14453:
> you apply this for the `filenameInfoFormat`, right?

Yes, but it can be used for `filenameFormat` too etc.

Nandaka:
if you used it in `filenameFormat`, will it duplicate the extension if you put it at the end?

Baa14453:
Something's not right it's only working sometimes i'll get back to this

PR #1311, opened 2023-11-30
Add deleteMembersByList and keepMembersByList

Terryble2:
Adds the option for a user to remove members and their images from their DB with a list. By default this option looks for a members.txt file in the main directory.

Could be extended with option to set members as deleted or only delete members and not their images.

Does not handle all edge cases, this PR is a working version which could be improved for safety and quality.

### DIFF
--- a/PixivDBManager.py
+++ b/PixivDBManager.py
@@ -14,6 +14,8 @@ from colorama import Back, Fore, Style
 import PixivHelper
 from PixivListItem import PixivListItem
 
+from PixivException import PixivException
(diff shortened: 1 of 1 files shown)


Terryble2:
I appended the new options as 14 & 15 to avoid user error but they would be more logically added as 12 & 13.

biggestsonicfan:
How often would you need to run this that you need a list? Wouldn't it be more logical to detect and handle multiple members in option 11?

Issue #1312, opened 2023-12-03
Ignore zip file from fanbox while re-encoding of all ugoira present in folder


HTYISABUG:
### Prerequisites

* [x] Did you read FAQ section in [readme.md](https://github.com/Nandaka/PixivUtil2/blob/master/readme.md)?
* [x] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?
* [x] Did you search for existing issues in [Issues](https://github.com/Nandaka/PixivUtil2/issues)?

### Description

When "re-encoding of all ugoira present in folder", script also processing zip files that are not ugoira.

Please ignore zip file from fanbox or add filter to filename such as `*ugoira*.zip` while "re-encoding of all ugoira present in folder", since creator usually provide file other than images or video such as PSD file using zip compression.

Also, ugoira should be a pixiv only feature ?

### Steps to Reproduce

1. Download a fanbox post that contents zip file
2. Run "re-encoding of all ugoira present in folder"

**Expected behavior:** Script only processes ugoira

**Actual behavior:** Script also processes files that are not ugoira

**Log file:** [Attach the `pixivutil.log` file in the application folder, recommended to delete the old file, reproduce the issue, and upload the newly generated file here]
[pixivutil.log](https://github.com/Nandaka/PixivUtil2/files/13538846/pixivutil.log)

### Versions
20230105

You can get this information from executing `PixivUtil2.py --help`.
Latest version available in https://github.com/Nandaka/PixivUtil2/releases

PR #1313, opened 2023-12-03
Fix re-encoding not ugoira zip files

HTYISABUG:
Possible fix issue #1312

### DIFF
--- a/PixivImageHandler.py
+++ b/PixivImageHandler.py
@@ -480,7 +480,7 @@ def process_ugoira_local(caller, config):
                 zip_name = os.path.splitext(os.path.basename(zip))[0]
                 zip_dir = os.path.dirname(zip)
                 image_id = zip_name.partition("_")[0]
-                if image_id not in list_done:
+                if 'ugoira' in zip_name and image_id not in list_done:
                     counter += 1
                     PixivHelper.print_and_log(None, f"# Ugoira {counter}")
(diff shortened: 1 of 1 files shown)


PixivImageHandler.py
```diff
@@ -480,7 +480,7 @@ def process_ugoira_local(caller, config):
                 zip_name = os.path.splitext(os.path.basename(zip))[0]
                 zip_dir = os.path.dirname(zip)
                 image_id = zip_name.partition("_")[0]
-                if image_id not in list_done:
+                if 'ugoira' in zip_name and image_id not in list_done:
```
Nandaka, line 483: this is assuming the zip file will always contains ugoira in the filename, right?

HTYISABUG: Yes.
According to my observation, all ugoira downloaded are in format

`{id}_ugoira{resolution} - {name}.zip`

But file attached in fanbox usually do not contain "ugoira" in filename, such as

`189281_p0_7RHGqTqoO0JzAJg23uGWPl9C - 過去絵-サイド高画質版-改.zip`.

FIlter by "ugoira" should be a acceptable method.

PatrickL546: I've checked my archive and there are no zip files with ugoira in the name. I have delete zip files set to True and I'm assuming the .ugoira files created from zip files have the same name so it should be safe. I hope

PR #1321, opened 2024-01-21
 Add option to manually specify file extension inside file names v2

Baa14453:
Version 2 of my previous PR #1266, sorry about that. Fixed the remaining issue, here's the original description:

Digikam recently changed the behaviour of XMP sidecar files and now they need to match the full file-name of the associated image, ie. what used to be:

    image.jpg
    image.xmp

is now:

    image.jpg
    image.jpg.xmp

This PR adds the option to use %image_ext% in the file naming process so I can change my filename config to say `image.%image_ext%`.

`%image_ext%` does not replace the existing extension, adding it to anything other than info files will just duplicate the image's file extension, but who knows maybe someone wants that somewhere, like making directories based on extension I don't know.

The new addition is a bit of a wild one, I really wasn't sure on the best way to handle Ugoiras.
The problem I ran into is that, at the time we create metadata files, we don't yet know what ugoira re-encodes we're going to create, and ugoira's are a bit unique in that the original file-type of a ZIP is not usually retained, at least by me anyway.

After a lot of umming and ahhing I decided to add a big If Statement and just rely on the config file options to create additional sidecars, but only where someone has enabled `writeImageXMPPerImage`. This config option currently creates additional XMP files for every image in an album, I think the method of creating multiple re-encodes of Ugoiras is similar enough behaviour that I've integrated it into this option.

Not enabling `writeImageXMPPerImage` will make it write just a single `xmp` sidecar still. If you use the new %image_ext% in your info filename format then it will be called `.zip.xmp`, not sure who wants that but maybe someone.

Finally, if you enable `writeImageXMPPerImage`, have Ugoira re-encodings enabled, but don't specify the `%image_ext%` in the Info filename format, then it will technically just write multiple XMP files all with the same filename, overwriting each one. I thought about adding some logic to avoid this to prevent the unnecessary writes but it's basically instant anyway so I didn't bother.

Please let me know what you think.

### Screenshots
Before:
image

After:
image


Baa14453:
:eyes:

readme.md
```diff
@@ -539,6 +539,7 @@ Please refer run with `--help` for latest information.
 - writeImageXMPPerImage
 
   Set to `True` to export the image information to a .XMP sidecar file, one per image in the album. The data contained within the file is the same but some software requires matching file names to detect the metadata. If set to `True`, then `writeImageXMP` is ignored.
+  Additionally, enabling this option will create a .XMP sidecar for every ugoira encoding enabled, and allow you to customise the name of each file using `%image_ext%`. For example, if you enable `createWebp` and `createGif`, then set your `filenameInfoFormat` to something like `%urlFilename%.%image_ext%`, then you will end up with `<image ID>.gif.xmp` and `<image ID>.webp.xmp` files created.
```
Nandaka, line 542: correct me if I'm wrong, but if `%image_ext%` not used in the filename format, when both ugoira and webm flag enabled for example, then it will only create 1 xmp file, right?

Baa14453: > correct me if I'm wrong, but if `%image_ext%` not used in the filename format, when both ugoira and webm flag enabled for example, then it will only create 1 xmp file, right?
That's correct, it will try to create multiple ones but they will just be overwritten.

PR #1356, opened 2024-07-22
Add support for encoding AVIF

0fbcb238c0:
This commit adds support for AVIF encoded Ugoiras.
The default settings are `-cpu-used 4 -crf 0 -row-mt 1 -tile-columns 2 -tile-rows 2 -vsync 0`

I chose lossless encoding because AVIF offers good filesize even with lossless, often being not much larger or even smaller than lossy WebP, while retaining the full image quality.


0fbcb238c0:
I have not added a parameter defining loops as AVIF loops by default. Should it still be added?

Nandaka:
> I have not added a parameter defining loops as AVIF loops by default. Should it still be added?

I think should be fine

PR #1373, opened 2024-09-14
ADD: Download from unlisted post by ID

OozoraHaruto:
Allows users to download unlisted posts from pixiv

As the page and APIs are similar to the artwork page, I decided to just edit the code a little and allow reused exiting codes to help with the downloading.

Let me know if I need to make any changes!

### DIFF
--- a/PixivBrowserFactory.py
+++ b/PixivBrowserFactory.py
@@ -386,7 +386,7 @@ class PixivBrowser(mechanize.Browser):
         parsed = ""
         if login_cookie is None or len(login_cookie) == 0:
             login_cookie = self._config.cookieFanbox
-        
+
         # Issue #1342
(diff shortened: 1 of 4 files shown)


Nandaka:
got sample for the unlisted post? never see that before?

OozoraHaruto:
I have added a link in the comments of how the URL looks like.
here it is "https://www.pixiv.net/artworks/unlisted/SbliQHtJS5MMu3elqDFZ", one probably needs to be whitelisted by the artist, I am unsure how it works. Here is a screenshot of how the page looks like.
スクリーンショット 2024-09-21 8 07 30

Nandaka:
noted, link is nsfw btw, so I edited it so it doesn't go directly

Issue #1348, opened 2024-07-03
Non-interactive new illust from bookmarked members doesn't work without explicit count limit


Hamuko:
### Prerequisites

* [x] Did you read FAQ section in [readme.md](https://github.com/Nandaka/PixivUtil2/blob/master/readme.md)?
* [x] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?
* [x] Did you search for existing issues in [Issues](https://github.com/Nandaka/PixivUtil2/issues)?

### Description

Non-interactive bookmark downloading (option 8) is broken in 20240703 as it tries to run `len()` on an integer, most likely `len(-1)` as that seems to be the default for `--bcl` / `--bookmark_count_limit`, which is not a valid operation in Python.

This can be worked around by adding the bookmark count limit explicitly (for example `--bcl=-1`), as that will supply the value as a string instead, but there's really no way to figure it out without reading the source code.

Most likely broken by #1324.

### Steps to Reproduce

1. Run PixivUtil2 with `-s 8 --ep=3 -x` arguments.

**Expected behavior:**

Three first bookmarked pages are downloaded.

```
Login successful.
My User Id: xxxxxxx.
Premium User: False.
End Page = 3
Processing New Illust from bookmark
Page #1
Image #1
Processing Image Id: xxxxxxxxx
[...]
Limit or last page reached.
Done.
```

**Actual behavior:**

PixivUtil2 crashes immediately after login.

```
Login successful.
My User Id: xxxxxxx.
Premium User: False.
End Page = 3
Traceback (most recent call last):
  File "/PixivUtil2.py", line 1745, in main
    np_is_valid, op_is_valid, selection = main_loop(ewd, op_is_valid, selection, np_is_valid, args, options)
  File "/PixivUtil2.py", line 1422, in main_loop
    menu_download_new_illust_from_bookmark(op_is_valid, args, options)
  File "/PixivUtil2.py", line 648, in menu_download_new_illust_from_bookmark
    if bookmark_count is not None and len(bookmark_count) > 0:
TypeError: object of type 'int' has no len()
Unknown Error, please check the log file: (<class 'TypeError'>, TypeError("object of type 'int' has no len()"), <traceback object at 0x7f22113ff080>)
```

### Versions

Broken in 20240703. Worked in 20230105.

[...]

PR #1378, opened 2024-09-29
Fixes Nandaka/PixivUtil2#1348 because it annoys me

KSuiz:


biggestsonicfan:
Why do you need to specify a bookmark limit of infinite when you shouldn't need to specify a limit at all if you don't want the limit used?

KSuiz:
Could also just set the default to something else but I don't really want to go around and make sure that won't break things so hacky fix ig

Issue #1403, opened 2025-02-04
Compatibility Issue with beautifulsoup4 4.13.1


marierose147:
**Description**:  
After reinstalling Python and dependencies, I encountered the following error when trying to run PixivUtil2 version `20250103` using `beautifulsoup4` version `4.13.1`:

```
TypeError: can't multiply sequence by non-int of type 'str'
```

This error occurred in the following stack trace:
```
Error at doLogin(): (<class 'TypeError'>, TypeError("can't multiply sequence by non-int of type 'str'"), <traceback object at 0x000001698590B700>)
Traceback (most recent call last):
  File "D:\PixivUtil2-master\PixivUtil2.py", line 1558, in doLogin
    result = __br__.loginUsingCookie()
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\PixivUtil2-master\PixivBrowserFactory.py", line 354, in loginUsingCookie
    parsed_str = str(parsed.decode('utf-8'))
                     ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\myusername\AppData\Local\Programs\Python\Python312\Lib\site-packages\bs4\__init__.py", line 1141, in decode
    return prefix + super(BeautifulSoup, self).decode(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\myusername\AppData\Local\Programs\Python\Python312\Lib\site-packages\bs4\element.py", line 2435, in decode
    piece = self._indent_string(
            ^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\myusername\AppData\Local\Programs\Python\Python312\Lib\site-packages\bs4\element.py", line 2520, in _indent_string
    space_before = formatter.indent * indent_level
                   ~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~
TypeError: can't multiply sequence by non-int of type 'str'

Cannot Login!
```

**Steps to Reproduce**:
1. Install Python 3.12.
2. Install dependencies as specified in `requirements.txt`.
3. Run the PixivUtil2 script (version `20250103`).
4. Encounter the error during login.

**Workaround/Fix**:
After downgrading `beautifulsoup4` to version `4.12.3`, the issue was resolved and the script runs as expected.

PR #1404, opened 2025-02-04
Fix unnecessary decode() calls on BeautifulSoup objects (#1403)

poyazi:
This PR fixes the error reported in Issue #1403 by removing unnecessary `.decode('utf-8')` calls when converting BeautifulSoup objects to strings. Specifically, all occurrences of:

```py
str(parsed.decode('utf-8'))
```

have been replaced with:

```py
str(parsed)
```

in the following functions:
- `loginUsingCookie()`
- `fanboxLoginUsingCookie()` (including Cloudflare challenge handling)
- `processLoginResult()`

This eliminates the `TypeError: can't multiply sequence by non-int of type 'str'` without requiring a downgrade of `beautifulsoup4`.


NHOrus:
Works for me

poyazi:
Thanks for testing!

Issue #1389, opened 2024-11-15
Incomplete download adds entry to database


biggestsonicfan:
### Prerequisites

* [X] Did you read FAQ section in [readme.md](https://github.com/Nandaka/PixivUtil2/blob/master/readme.md)?
* [X] Did you test with the latest [releases](https://github.com/Nandaka/PixivUtil2/releases "latest releases") or [commit](https://github.com/Nandaka/PixivUtil2/commits/master "latest commit") ?
* [X] Did you search for existing issues in [Issues](https://github.com/Nandaka/PixivUtil2/issues)?

### Description

It looks like there may be a race condition where an incomplete download will still write to the database.

### Steps to Reproduce

1. Try to download image id 121384542
2. Download should fail as incomplete
3. Verify image is in database with the database manager and option 6.

**Expected behavior:**

Download should succeed or not write to database on failure.

**Actual behavior:**

Download is incomplete and fails, yet the entry is added to the database.

**Log file:** 
[pixivutil.log](https://github.com/user-attachments/files/17777130/pixivutil.log)

### Versions
Latest git pull at 21f7bede602d83a97617cf8b313ba0eac59f59ad (PixivUtil2 20240703)

Litoba:
I'm having same problem.
![캡처_2024_11_24_01_23_55_407](https://github.com/user-attachments/assets/cd1fa3e3-9e24-438b-a65b-3a24930038ec)

missateneis:
Hopefully this problem gets a fix. There is no way to know about incomplete downloads unless you catch one by accident. alwaysCheckFileSize option requires all files to be present in folder and takes long time to process.

PR #1409, opened 2025-03-17
Fix incomplete downloads not recognized as such

RadCraftplay:
PixivUtil2 deletes partially downloaded files, but doesn't try to download them again. On top of that it assumes these images were downloaded completely, marking them as such in the database (issue #1389).

The problem was that in ``PixivHelper.download_image``, when downloading file in parts, if Pixiv didn't send all requested parts (and therefore ``curr == prev``), PixivUtil2 would assume the download has finished, but the file size was not known: 

```python
            # check if downloaded file is complete
            if file_size > 0 and curr == file_size:
                total_time = (datetime.now() - start_time).total_seconds()
                print_and_log(None, f' Completed in {Fore.CYAN}{total_time}{Style.RESET_ALL}s ({Fore.RED}{speed_in_str(file_size, total_time)}{Style.RESET_ALL})')
                break

           elif curr == prev:  # no file size info
                total_time = (datetime.now() - start_time).total_seconds()
                print_and_log(None, f' Completed in {Fore.CYAN}{total_time}{Style.RESET_ALL}s ({Fore.RED}{speed_in_str(curr, total_time)}{Style.RESET_ALL})')
                break
```

In that case, ``file_size`` variable has to be checked. When file size is not known, the variable ``file_size`` is set to ``-1``. There is a third case, when file size is provided, but no further parts are provided. The code had to be updated:

```python
            # check if downloaded file is complete
            if file_size > 0 and curr == file_size:
                total_time = (datetime.now() - start_time).total_seconds()
                print_and_log(None, f' Completed in {Fore.CYAN}{total_time}{Style.RESET_ALL}s ({Fore.RED}{speed_in_str(file_size, total_time)}{Style.RESET_ALL})')
                break

            # no file size info
            elif file_size < 0 and curr == prev:
                total_time = (datetime.now() - start_time).total_seconds()
                print_and_log(None, f' Completed in {Fore.CYAN}{total_time}{Style.RESET_ALL}s ({Fore.RED}{speed_in_str(curr, total_time)}{Style.RESET_ALL})')
                break

            # incomplete download
            elif file_size >= 0 and curr == prev:
                raise PixivException(f"Download incomplete for: {url}", errorCode=PixivException.DOWNLOAD_FAILED_OTHER)
```

Even though in the ``finally`` block there was a condition that would check is the file size is invalid, it would not fail the download. By raising exception in the loop, PixivUtil2 acknowledges the error, waits couple of seconds, and retries just like in case of any other error.

# Possible issues

I'm not sure if using ``PixivError`` in that case is correct, but it seems fine. In the block that handles the exception, the same exception and error code is set.

This change makes PixivUtil2 exit with non-zero code. This may break some scripts that expect the program to exit with no errors. Also, every time the download fails, the stack trace is printed, which might confuse the user.

Another thing is that when the file size is not known, there is no way to check whether the download is complete or not. Even though that sounds bad, Pixiv doesn't seem to send partial data when the file size is not known (or at least I didn't notice having partially downloaded images).

### DIFF
--- a/PixivHelper.py
+++ b/PixivHelper.py
@@ -811,10 +811,15 @@ def download_image(url, filename, res, file_size, overwrite):
                 print_and_log(None, f' Completed in {Fore.CYAN}{total_time}{Style.RESET_ALL}s ({Fore.RED}{speed_in_str(file_size, total_time)}{Style.RESET_ALL})')
                 break
 
-            elif curr == prev:  # no file size info
+            # no file size info
+            elif file_size < 0 and curr == prev:
                 total_time = (datetime.now() - start_time).total_seconds()
                 print_and_log(None, f' Completed in {Fore.CYAN}{total_time}{Style.RESET_ALL}s ({Fore.RED}{speed_in_str(curr, total_time)}{Style.RESET_ALL})')
                 break
+
+            # incomplete download
+            elif file_size >= 0 and curr == prev:
+                raise PixivException(f"Download incomplete for: {url}", errorCode=PixivException.DOWNLOAD_FAILED_OTHER)
 
             prev = curr
     except OSError as ex:


Ikkoru:
Do you know how long this has been going on for, and if there's a way to only download the incomplete files? Or will I have to redownload all missing files?

Ikkoru:
The bigger problem for me is that I have manually deleted a large number of files due to having higher quality versions.
I guess I'll just have to redownload everything from the past half year or so, and manually delete the lesser quality versions that get downloaded again... ;  ;

Issue #1421, opened 2025-04-17
Fail to get the imageID


RMWW:
There is a new problem again. It prompts that the imageID cannot be obtained.

![Image](https://github.com/user-attachments/assets/703cc431-24ef-4cbf-b880-1694f65e6a3d)

Fulgur14:
For me, this started to happen in the middle of download. And when I restarted the program, it could no longer log in, and hasn't worked since then, so it might be related?

rts236:
I had a similar problem that solved itself by setting `r18mode` to `True`, though it may not be related since I'm using a frankenstein version (mostly from dkhex).

Fulgur14:
> I had a similar problem that solved itself by setting `r18mode` to `True` winthin `config.ini`, though it may not be related since I'm using a frankenstein version (mostly from dkhex).

Hm, tried that, but doesn't work for me.

rts236:
What version are you using? I haven't been able to get this Nandaka repository to work, but dkhex's `fixes` branch has a working version for individual images. My guess is that dkhex doesn't want to push something that does not fully work on the main repository, thus having it on their fork. https://github.com/dkhex/PixivUtil2/tree/fixes
I modified their version for it to also work with `member_ids`, but the problem is that some features (Pixiv-Fanbox unique user name) cannot solved (remember to enable `r18mode` or errors may happen). https://github.com/rts236/PixivUtil2/tree/fixes
Put a merge request so you can check the differences with dkhex. https://github.com/dkhex/PixivUtil2/pull/1
The thing is that I don't know if other features are working, I only use the single image and member_ids options. Because of the broken feature of my version, cross-compatibility between Pixiv and Fanbox may not function.

Fulgur14:
[...]

Issue #1414, opened 2025-04-01
Failing all image downloads


wmjdgla:
Started happening on Monday, was working the day before despite no change in config.
I'm on the latest master commit.
Able to login, but as any attempt to download fails with:
```
Image ids: 128628526
Processing Image Id: 128628526 
Image ID (128628526): 'NoneType' object has no attribute 'artist' 
Dumping html to: Error medium page for image 128628526.html 
Stack Trace: (<class 'AttributeError'>, AttributeError("'NoneType' object has no attribute 'artist'"), <traceback object at 0x000001DB8CF8CB80>) 
Traceback (most recent call last):
  File "d:\Dev\PixivUtil2\PixivImageHandler.py", line 97, in process_image
    caller.set_console_title(f"MemberId: {image.artist.artistId} ImageId: {image.imageId}")
                                          ^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'artist'
```
Seems like problem is with this line:
https://github.com/Nandaka/PixivUtil2/blob/c61f3e225c39a8b460b34a87cf51e787f9530009/PixivBrowserFactory.py#L662
I stepped into the debugger and it looks like `response` no longer contains any "meta-preload-data" string, so the entire `if` block is skipped and `getImagePage` returns `None` for `image`, thereby causing the exception. I tried commenting away the conditional check to see if the code block can work without it, but no luck. Seems like Pixiv changed their entire response payload? Web stuff is totally out of my depth, so this is as far as I can go.

dkhex:
Got the same error.
Request to https://www.pixiv.net/artworks/{image_id} no longer have artist information.
I made workaround by making request to https://www.pixiv.net/ajax/illust/{image_id} and work with a json, instead of parsing html.
It works for my usecase (saving bookmarks), but I pretty sure it will break something.
https://github.com/dkhex/PixivUtil2/tree/fixes
I didn't dive deep enough into codebase, so I need more time for proper fix.

dioporcos:
Image ID (xxxxxxx): 'NoneType' object has no attribute 'imageId'
Dumping html to: Error medium page for image xxxxxxx.html
Stack Trace: (<class 'AttributeError'>, AttributeError("'NoneType' object has no attribute 'imageId'"), <traceback object at 0x000001CC09928300>)

getting this error

XiaokeKQ:
Solved your code?

cocokakao:
same issue

[...]

PR #1423, opened 2025-04-19
Pixiv artist and image information source change (#1414 #1421)

rts236:
@dkhex:
* Move to get artist and image information from ajax JSON.

@rts236:
* Patch `member_ids` operation and reduce/comment deprecated/conflicting code with new Pixiv format.


rts236:
For the love of what is sacred, I cannot find how to smush the commits into a single one before the pull request. I don't wanna push 8 commits with random tidbits, but I don't find any way to do it from the browser UI. 

Is it because it has @dkhex commits? Is this something that can be done after the pull request? This is worse than the Python thing, and I just refuse to use git desktop. Nah, I'll have to give in tomorrow.

NHOrus:
I do not think github ui allows that. You'd want cli or some non-gihub git clients and gorce push.

rts236:
Okey, had to rebuild repository because it wouldn't let me squash at my repository.
The thing I don't understand is why it then failed to merge without conflicts, but hey, solved the conflicts by forcing what was from my repo and now is ready for merge. Hope I didn't break anything.

Nandaka:
Ah I think I also do the same thing as you do in parallel

PixivArtist.py
```diff
@@ -25,32 +25,22 @@ class PixivArtist:
...
         self.artistId = mid
-
-        if page is not None and len(page) > 0:
-            payload = None
+        
+        if payload_body is not None:
             # detect if image count != 0
             if not fromImage:
-                payload = json.loads(page)
-                if payload["error"]:
```
Nandaka, line 38: need to add error check here

Nandaka:
also, can you check if it pass the unit test

PixivArtist.py
rts236: Are you sure? This is already the `body` of the JSON, so there should not be an `error` field anymore since that is at the same level as the `body`. Or am I wrong? I'm have no experience on this.
This comment reviews every place where an `PixivArtist` object is created, and in both I found (hope didn't miss any) the `error` field is checked previous to generating the `PixivArtist` object. 
https://github.com/Nandaka/PixivUtil2/issues/1421#issuecomment-2816697082

rts236:
I'll need more information about `unit test`. Just realized that there're `test_*.py` files, and running them show errors. 
Whatever it may be, I realized that after the pull I made on my repository now it doesn't work, and I don't know why. I'm reviewing differences and will try to push something.

rts236:
Now my branch works. The tests are still giving errors, but downloading image and members work. However, I believe that if I merge last commits from this repository to mine (that what's the `This branch has conflicts that must be resolved` I believe is meaning, then it stops working.

What a mess is working with github, or I'm just that bad at it.

PixivArtist.py
Nandaka: ah, yeah, I notice you only pass the body part. Then the checking need to be done on each caller before passing the body part ==> that's why I just pass the json string to this function originally, and do all the validation here once.

rts236:
I can confirm that, after rolling back some of your changes (sorry), now it's back working for both images and member_ids. 
The tests.py are still giving errors, but in some they still use OAuth, so I'm not sure if they should be updated.

PixivArtist.py
rts236: Looking by what was done previously, I believe they changed the check to be done at source of the call. I guess that way makes it easier to have different error calls for every use without having to put more filter options in the function. 
Different ways of doing things have different positive and negatives I guess.

rts236:
Looking into the tests, they seem to differ a lot from the current implementation, so let me summarize some of the changes (not only mine) that affect some of the tests. This is the current workflow for a member:
  * Get ajax JSON from `https://www.pixiv.net/ajax/user/1813972/profile/all`.
  * Generate a `PixivArtist()` object in order to take the list of `illusts`, `manga` and `novel` [`imageList`].
  * Get the first entry from the list for getting rcp JSON from `https://www.pixiv.net/rpc/get_work.php?id=1039353`.
    * If the artist has no matching art with tags (r18mode included) or doesn't have any art altogether, return to main menu with exception. We don't have any way of getting full artist information, as far I know, thus this trap.
  * Pass that payload to `ParseInfo()` which will take most artist data [`artistId`, `artistToken`, `artistName`]
  * Last but not least, get ajax JSON from `https://www.pixiv.net/ajax/user/1813972` in order to get [`artistAvatar`, `artistBackground`].
For straight an image info, the creation of a `PixivArtist()` object calls `ParseInfo()`, that then calls for `ParseInfoFromImage()` using only `https://www.pixiv.net/ajax/illust/1039353` ajax JSON payload. Then it doesn't get more info because it doesn't need  [`imageList`, `artistAvatar`, `artistBackground`]. Why use rcp instead of ajax with members? I assume that it's for treating different errors, but I really don't know. The filters are done for each specific case at `PixivArtist.py` (that has been most of my work, but it hasn't been me alone).

rts236:
Oh, realized you guys solved the problem on your own. Then I'll be closing the pull request, seeing that your code is more reliable than mine (easier to trace down the commits and not something, at least). Feel free to take any indications or part of my code though (maybe not including dkhex's)!

Issue #1416, opened 2025-04-02
Download Pixiv galleries as a ZIP archive


psilabs-dev:
Fork issue: https://github.com/psilabs-dev/PixivUtil2/issues/10

Feature: Support the option to download Pixiv galleries as ZIP (or CBZ...) archives.

My current workflow is to download Pixiv galleries with PixivUtil2, then perform automated uploads as zip archives to a comic server like LANraragi for viewing. This is because LANrargi mainly supports reading from archives. Therefore, implementing this feature allow LRR to just read the downloaded files, effectively halving the local Pixiv file size.

On design: for each illust ID, create a temporary staging directory and direct Pixiv images there. Once done, put them into a zip archive in that temp directory. On success, move the zip archive to the target save path. This way, we will minimize incomplete or corrupted archives/images.

I'll be working towards implementing this feature, as I will be need for this. But if this is also something wanted upstream, I would be happy to receive feedback and make a pull request, though getting to this may take a while.

PR #1446, opened 2025-05-11
Support archive mode for Pixiv artworks

psilabs-dev:
Issue: #1416

Allows users to download Pixiv artwork images into an archive using zipfile, instead of to a directory.

To use this, set `createPixivArchive = True`. This will by default create uncompressed archives out of Pixiv artworks. No changes to `filenameformat` are required: the zip archive will be inferred from the leaf directory and given a ".zip" extension.

Customizations on compression are made available by `createPixivArchiveCompressionType` and `createPixivArchiveCompressionLevel`.

Since I don't use Fanbox, this implementation is limited to Pixiv downloads only.

As discussed in the issue, images are first saved to a temp directory, then put into a temp archive, which is finally moved to a persistent archive. This minimizes room for interruptions and partial downloads, ensuring that only completed archives make it to the final destination.

An important archive-mode change to database insertion for images: image save_names in `pixiv_manga_image` table will be the base names of the filenameformat. For example, if in directory mode the image has save name "a/b/c.png", then in archive mode the artwork in `pixiv_master_image` will have save name "a/b.zip" and the image in `pixiv_master_image` will have save name "c.png". I think this is a reasonable interpretation of save name, as it is the image's position within the archive.

I thought about writing a migration script, but considering all the possible configurations, it is probably better not to give a false sense of security...

### DIFF
--- a/PixivConfig.py
+++ b/PixivConfig.py
@@ -141,6 +141,10 @@ class PixivConfig():
         ConfigItem("Filename", "tagTranslationLocale", "en"),
         ConfigItem("Filename", "customBadChars", "", followup=PixivHelper.parse_custom_sanitizer),
         ConfigItem("Filename", "customCleanUpRe", "", followup=PixivHelper.parse_custom_clean_up_re),
+        ConfigItem("Filename", "createPixivArchive", False),
+        ConfigItem("Filename", "createPixivArchiveCompressionType", "ZIP_STORED",
+                   restriction=lambda algorithm: algorithm in {"ZIP_STORED", "ZIP_DEFLATED", "ZIP_BZIP2", "ZIP_LZMA"}),
+        ConfigItem("Filename", "createPixivArchiveCompressionLevel", 0, restriction=lambda level: level in range(0, 10)),
 
         ConfigItem("Authentication", "username", ""),
         ConfigItem("Authentication", "password", ""),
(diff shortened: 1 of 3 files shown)


psilabs-dev:
Code should be stable now, please review/try it out and provide feedback if you have a chance.

Although the actual archival code is simple, an attempt was made to handle many possible scenarios; all of which arise in the case of migration. However, as seen by the commit history, unexpected bugs can still occur. Despite having logic to handle migration and certain levels of abuse, my recommendation for users is to either start over, or (if experienced) do a manual migration for their specific scenario, with plenty of backing up and data validation.

I spent some time figuring out how to test this, however I don't think tests would be too helpful with the current monolithic function design. There would be a large amount of config, db and caller mocking done, such that the resulting tests do not give confidence to correct logic. Perhaps a better test designer than me can address this.

This has been manually tested on macos/linux (not tested on windows) and with docker (conda image), and here is a screenshot demo of archive mode in a downloader microservice with pixivutil engine.

demo

An issue is that it is not obvious what format an ugoira's save name should be in the database, when the user keeps only derivatives like gifs or mp4. During a database migration, if the file is saved as a ".gif" file instead of ".zip", ffmpeg may throw a 254 error on a second pass.

Also the amount of case handling and complexity added to PixivImageHandler might be undesirable for the project, so I understand if we don't want to support this feature. Currently this implementation is sufficient for my use case.

PR #1451, opened 2025-05-24
Removed (some) deprecation warnings, unclosed resource errors.

psilabs-dev:
Used resource context managers, updated some bs4 methods, made db testable. Added mocking for member so that some tests pass. Did not touch fanbox.

Btw are we still using all of these tests at the moment? The tests are failing on master.


Nandaka:
some test are deprecated because pixiv uses different layout (last time I use BeautifulSoup4 to parse the page), now it should be using pixiv REST api. I just lazy to remove it :P

PR #1481, opened 2025-11-08
Skip AI with cache when download by artist

psilabs-dev:
Adds a cache table to pre-emptively skip making repeated API calls to AI-tagged artworks.

If config.aiDisplayFewer is set and user is downloading Pixiv artworks by user, PixivUtil2 will check the local DB to see if the artwork ID's AI type data exists, and perform skip before API call is made. If no record exists, it will store the artwork's AI type data after metadata fetch, to avoid making an API call next time.

See https://github.com/psilabs-dev/PixivUtil2/issues/22 for downstream issue/use case. Though this does involve creating a new table just to skip, so I can understand if it's not worth it but feel free to leave feedback.

### DIFF
--- a/PixivDBManager.py
+++ b/PixivDBManager.py
@@ -127,6 +127,16 @@ class PixivDBManager(object):
                             last_update_date DATE,
                             PRIMARY KEY (image_id, tag_id)
                             )""")
+            
+            # image ID is primary key, may not reference to pixiv_master_image as it may not
+            # be downloaded. Used for filtering out AI images.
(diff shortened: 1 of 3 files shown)


psilabs-dev:
Did a quick check, unless `(artist, list_page) = PixivBrowserFactory.getBrowser().getMemberPage ...` returns info about whether an artwork includes AI flag, then we don't have enough info from this response alone, since `imageList` is a list of image IDs, not image data.

PR #1496, opened 2025-12-09
Add database migrations

RadCraftplay:
Maybe the project is not updated often, but it is annoying that every single update I have to delete the database and config. So I made that at least you can keep the database. This will also help people who are updating PixivUtil2 via git.

If you are developer, every time the database is changed, you have to create new `.sql` file in `migrations` directory (for example: `002_ai_info.sql`). The template for the filename is `XXX_short_description.sql`, where XXX is incremental migration number (it decides the order of migrations) and short description tells us what has changed in the migration.

Before this branch is merged I could add some python script to help users that have already used PixivUtil2 before to mark migrations as done.

### DIFF
--- a/PixivDBManager.py
+++ b/PixivDBManager.py
@@ -36,177 +36,64 @@ class PixivDBManager(object):
     def close(self):
         self.conn.close()
 
+    def loadMigratedFiles(cursor):
+        try:
+            cursor.execute("SELECT file_name FROM migrations ORDER BY file_name")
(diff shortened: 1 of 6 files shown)


RadCraftplay:
> > but it is annoying that every single update I have to delete the database and config.
> 
> What? Why? No. You don't need to do this. Ever. PixivUtil2 is very good about preserving existing database layout. And rewriting config to correct version.

My bad

PR #1506, opened 2026-02-08
GUI

Pgj1997:
I've created a GUI for the program.

This might necessitate the people who were complaining about there not being an EXE anymore


Pgj1997:
Okay, so you guys don't like the readme I added.

Understood.

PR #1507, opened 2026-02-09
Add command options for metadata downloading by ID (artist, image, tag, series)

psilabs-dev:
As time progresses, metadata of artworks may be outdated, or incomplete due to introduction of new features (e.g. saving by caption, tag, series), and re-downloading all content including images may not be necessary.

This PR/feature spec aims adds options to forcefully update metadata by ID, through additional command line entrypoints m1-m4. Details below.

Commands (Pixiv only):

- m1: download member metadata by ID
- m2: download image metadata by ID
- m3: download series metadata by ID
- m4: download tag metadata by ID

For all table fields, coalesce, update, and default generally follow these rules: the "freshest" data is saved to database.

- API response if not null always overrides.
- If API response field null, database field value if not null shall not be overwritten.
- If both API and database fields are null, apply default if default is specified.

## download artist metadata by ID

This will download and update the `pixiv_master_member` row only. While other member update methods may or may not update, this method should *always* update the corresponding row.

("default" means: update if field does not exist with default value)

- member_id -> update
- name -> update
- save_folder -> update to `N\A` (not saving content, so this is enforced)
- created_date -> coalesce default now
- last_update_date -> update (to now)
- last_image -> coalesce, update or default -1
- is_deleted -> coalesce, update or default 0
- member_token -> coalesce, update or default null

E.g.: update member info by ID -> makes ajax call to member info. If member info exists, fetches member info from db, then applies update with coalesce

Unlike the content download option "1", download artist metadata by ID *only* downloads the metadata of the artist, not images posted by the artist. However, we may have an argument for including image metadata (and corresponding tags/series) owned by artist. This is currently not planned.

configs blocking this (e.g. autoAddMember) would be ignored. Otherwise m1 has no point.

## download image metadata by ID

Download and update image, member, series, and tag tables + mapping tables. This follows the existing image handler logic, but should always update, so check for `insert or ignore`.

Image/tag metadata updates are overwriting, not additive. That means: remove all existing image/tag associations by this image ID, and re-apply image/tag mappings. Also, remove image/series associations, and re-apply image/series mappings.

For example:

- if image metadata has no series, check if image is part of series, and remove mappings.
- if image metadata has series, check if image mapping is correct, else update series metadata (see download series metadata by ID).
- always overwrite image/tag associations: delete existing mappings, then re-apply mappings corresponding to current image tag.
- always update artist info with coalesce or default null.

An important point: because download image metadata by ID does not download images, `pixiv_manga_image` shall NOT be updated.

An m2-unique note: Metadata download types are still subject to config flags. For example, caption data will never be downloaded even through this entrypoint, as long as config doesn't permit it. Same applies to auto add series/tag/member.

Config flags are prioritized over freshness. Put another way, config flags gover whether a field can be written to (updated/deleted). If config flag is disabled for an extra field, then the corresponding field may not be touched.

An exception applies if this is existing behavior (e.g. captions). In the case of captions, disabling autoAddCaption would overwrite existing captions with empty string. This is one of those cases where I think: we shouldn't remove captions if autoAddCaption is enabled.. May require further discussion.

## download series metadata by ID

Download and update `pixiv_master_series` and `pixiv_image_to_series_map". This should delete existing image/series associations, and update the map table with new image/series associations, even if the image metadata does not exist in `pixiv_master_image` or is saved locally.

Metadata for image does not need to be fetched.

- series_id -> update
- series_title -> update
- series_type -> update
- series_description -> update
- created_date -> coalesce or default now
- last_update_date -> update

This is not affected by config auto add series flag.

## download tag metadata by ID

Download and update `pixiv_master_tag` and `pixiv_tag_translation`. Do NOT delete existing tag translations; if a translation is missing, keep old translation. However, new translations of the same language override the current translation in database.

`pixiv_master_tag`:
- tag_id -> same
- created_date -> coalesce or default now
- last_update_date -> update

`pixiv_tag_translation`:
- tag_id -> same
- translation_type -> same
- translation -> update
- created_date -> coalesce or default now
- last_update_date -> update

Note: Pixiv allows any string (at least afaik) to be a tag. Therefore, an option exists to only save tags which have a pixpedia entry/have translations. However, know that not all recognized tags may have these entries; therefore, use "m4" responsibly.

This is not affected by config auto add tag flag.

---

While I do not use the command line, I intend to apply these changes eventually for downstream use by PixivUtil server, as metadata management becomes an increasing downstream concern

This PR has a bit of business logic complexity/possible inconsistency, as I'm not sure whether to rewrite metadata-specific logic (which would introduce a lot of possibly duplicate code hard to refactor), or to rely on existing db logic (which has quirky behavior users may expect at this point). A mix of both exists but hopefully the spec described is respected.

### DIFF
--- a/handler/PixivImageHandler.py
+++ b/handler/PixivImageHandler.py
@@ -642,7 +642,10 @@ def process_image(caller,
 
             if metadata_only:
                 filename = r[0] if in_db else 'N/A'
-            db.updateImage(image.imageId, image.imageTitle, filename, image.imageMode)
+
+            # Only update caption when autoAddCaption is enabled; otherwise preserve existing value.
+            caption_to_update = image.imageCaption if config.autoAddCaption else None
+            db.updateImage(image.imageId, image.imageTitle, filename, image.imageMode, caption=caption_to_update)
 
             if len(manga_files) > 0:
                 if archive_mode_update_manga_image_paths:


psilabs-dev:
Dogfed this a bit (went ahead and updated all my old artworks without a caption), no apparent logic errors, feel free to test it out/review when you get a chance thanks

Issue #1538, opened 2026-07-13
[Fanbox] KeyError('creatorId') when trying to download this post


MarqFJA87:
Getting the following error message whenever I try to download this one particular post (**NSFW**); other Fanbox posts from other users were downloaded just fine.

```
Input: f3
Post ids = 12244545
Trying to log in FANBOX with saved cookie
FANBOX Login successful.
Traceback (most recent call last):
  File "D:\Programs\Bulk downloaders\PixivUtil2\PixivUtil2.py", line 1768, in main
    np_is_valid, op_is_valid, selection = main_loop(ewd, op_is_valid, selection, np_is_valid, args, options)
                                          ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Programs\Bulk downloaders\PixivUtil2\PixivUtil2.py", line 1500, in main_loop
    menu_fanbox_download_by_post_id(op_is_valid, args, options)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\Programs\Bulk downloaders\PixivUtil2\PixivUtil2.py", line 998, in menu_fanbox_download_by_post_id
    post = __br__.fanboxGetPostById(post_id)
  File "D:\Programs\Bulk downloaders\PixivUtil2\common\PixivBrowserFactory.py", line 1103, in fanboxGetPostById
    artist = self.fanboxGetArtistById(js["body"]["creatorId"])
                                      ~~~~~~~~~~^^^^^^^^^^^^^
KeyError: 'creatorId'
Unknown Error, please check the log file: (<class 'KeyError'>, KeyError('creatorId'), <traceback object at 0x000001CBFCBF7740>)
```

The log is attached below.

[pixivutil.log](https://github.com/user-attachments/files/29964387/pixivutil.log)

aksskl:
I'm getting three different errors trying to download from fanbox depending on how I try to download. By postid (f3), creatorid (f2 f5), or attempting to get supporting/following artists (f1 f4). `KeyError('creatorId')` as shown in the first post, `KeyError('title')`, and `TypeError('string indices must be integers') ` errors respectively.

```
Post  = 12094725
Title = pixivFANBOX個別規約を改定しました（2026年7月6日改定）
Type  = image
[...]

PR #1542, opened 2026-07-26
Implement download of latest supported FANBOX posts

xWTF:
## Problem

The existing FANBOX download modes process posts creator by creator or require individual post IDs. There is no convenient way to download the latest posts across all currently supported creators.

Specifically in my situation, where there're 40 supported creators, a single `f1` with 1 page, without any downloaded posts (update check only) would take 12 ~ 20 min to finish, and can be easily interrupted if there's a temporary network problem. I've spend way too much of time fixing my network and retry the f1 option in recent years.

## Solution

FANBOX provides the `post.listSupporting` endpoint for the update check purpose.

Added a new `f7` action to download the latest posts from supported FANBOX creators.

The new action:
- Fetches 10 posts per page from `post.listSupporting`. Technically you can set the page to 100 or whatever, but I choose to respect the web behaviour, in case they added another check on that later. Pagination is implemented via `body.nextUrl`
- Fetches the supporting creator list from `plan.listSupporting`. Although I'm not sure, the "Only show Creators you're currently supporting" toggle appears to be frontend only check, so I added this double check, just to make sure we didn't include any following-only creator by accident. It filters out posts whose `creatorId` is not in the supporting creator list.
- Completes the creator metadata by getting each creator, and use that on creation of post object

This PR also includes fix for `fanboxGetPostJsonById` and `fanboxGetPostsFromArtist`, as mentioned in #1538.

It also includes a QoL patch that constructs cookie for `fanboxGetPostJsonById` automatically, so you don't have to manually set the `cookieFanboxTemp` in `config.ini` for fanbox download to work.

The README and command-line help have been updated with the new `f7` action. I'm not sure why the `f6` is not included tho.

## Testing

Manually tested a few times and it appears to be working fine with 5 pages.

### DIFF
--- a/PixivUtil2.py
+++ b/PixivUtil2.py
@@ -183,6 +183,7 @@ def menu():
     print(' f4. Download from following list (FANBOX)')
     print(' f5. Download from custom list (FANBOX)')
     print(' f6. Download Pixiv by FANBOX Artist ID')
+    print(' f7. Download from latest supporting posts (FANBOX)')
     print(Style.BRIGHT + '── Sketch '.ljust(PADDING, "─") + Style.RESET_ALL)
     print(' s1. Download by creator id (Sketch)')
     print(' s2. Download by post id (Sketch)')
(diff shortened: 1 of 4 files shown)


Nandaka:
this one ready to merge?

xWTF:
> this one ready to merge?

Yes it's ready for review & merge. Sorry for the the confusing force pushes, made some mistake while staging changes (I have too many local changes and have to pick out relevant changes)

SimonPilkington:
@xWTF Thanks for this, very useful. Can you please explain what the `post.listSupporting` endpoint does exactly? I assume returns posts in chronological order but it's unclear what I get when I request 5 pages from `f7` which makes me worried about missing posts.

Is there a page on the www.fanbox.cc web that uses this endpoint? I thought it'd be the homepage but it uses `post.listHome`, and I failed to find any page using `post.listSupporting` by a cursory search.

xWTF:
> @xWTF Thanks for this, very useful. Can you please explain what the `post.listSupporting` endpoint does exactly? I assume returns posts in chronological order but it's unclear what I get when I request 5 pages from `f7` which makes me worried about missing posts.
> 
> Is there a page on the [www.fanbox.cc](http://www.fanbox.cc) web that uses this endpoint? I thought it'd be the homepage but it uses `post.listHome`, and I failed to find any page using `post.listSupporting` by a cursory search.

This endpoint is what **Posts** section of fanbox homepage uses:
image

When you turn on "Only show Creators you're currently supporting" there's no extra requests sent so I assume it's a frontend filtering (to strip out following-only creators), and hence the `plan.listSupporting` extra check.

There's no official guarantee that this Posts section shows posts in reverse chronological ordering, but per my observation it is, and this design makes sense. One of the evidence is pagination is done via `?maxPublishedDatetime=xxx&maxId=yyy` format, I guess the db is like `select stuff from posts where ts < $maxPublishedDatetime and id < maxId order by ts desc`.

I can't guarantee that this won't miss posts neither, but I'd like to give it some trust and use it. It significantly reduced my auto download task time ¯\\\_ (ツ)\_/¯

xWTF:
> @xWTF Thanks for this, very useful. Can you please explain what the `post.listSupporting` endpoint does exactly? I assume returns posts in chronological order but it's unclear what I get when I request 5 pages from `f7` which makes me worried about missing posts.
> 
> Is there a page on the [www.fanbox.cc](http://www.fanbox.cc) web that uses this endpoint? I thought it'd be the homepage but it uses `post.listHome`, and I failed to find any page using `post.listSupporting` by a cursory search.

Ah, I made an mistake - you're right about the frontend uses `post.listHome`, which is the endpoint that provides content of `Posts` page. The frontend fetches `post.listSupporting` on load simultaneously, and shows content of that when you toggle on the switch, this explains why there's no extra requests sent.

So I assume the double check with `plan.listSupporting` is unnecessuary now. (it won't break stuff though, just added one unnecessuary request)

SimonPilkington:
> This endpoint is what **Posts** section of fanbox homepage uses:
> image

I missed the only show supported toggle, my mistake. With that on, the page does indeed switch to using `post.listSupporting`. That clears everything up. Thanks again.

